Issue #37, opened 2015-01-11
Missing styles


bezoerb:
Hey @pocketjoso,

our tests for `critical` and `grunt-critical` are failing since your last release (0.2.52)
With 0.2.51 everything works fine.
I've created a gist with the resulting css before and after the version bump:
https://gist.github.com/bezoerb/04aa464eeecead2f7b0b

pocketjoso:
Thanks, sorry about this. Will sort out as soon as I find some time, hopefully tonight.

pocketjoso:
@bezoerb Sorry for delay, now resolved by #38 .

bezoerb:
@pocketjoso thanks!

Issue #36, opened 2015-01-08
CSS3 selectors (::selection) result in malformed CSS


optimalisatie:
CSS3 selectors are causing malformed CSS.

The following source code (part of **all.css**)

```css
::-moz-selection {
  color: #fff;
  background-color: #FE1486
}
::selection {
  color: #fff;
  background-color: #FE1486
}
```

Becomes the following after [grunt-penthouse](https://github.com/pocketjoso/penthouse):

```css
::-moz-selection {
  color: #fff;
  background-color: #FE1486
}
::selection {
  cul{
    list-style: none
  }
  .clear {
    clear: both!important
  }
  footer,header,nav {
    display: block
  }
  :focus {
    outline: 0
  }
  input {
    border: 1px solid #d8d5d5;
    background: #fff;
    padding: 5px
  }
}
```

The element **cul** is unknown (probably malformed code). When the CSS code is minified via cssmin it is further malformed. The bracket of **::selection {** does not have an ending anymore, causing a large part of the CSS file to be removed.

pocketjoso:
Thanks for reporting this bug - will fix it as soon as I have a chance.

optimalisatie:
Thanks a lot for the great plugin! I am looking forward to the update!

pocketjoso:
I can't reproduce your issue - just running the css you pasted here in `penthouse` generates the expected CSS. Please check what css is needed to cause the issue to appear, and let me know. Thanks

pocketjoso:
@optimalisatie Any update? Need to be able to reproduce before I can fix the issue.

optimalisatie:
Hi! I've managed to work around the problem by beautifying the full CSS code via [grunt-cssbeautifier](https://github.com/douzi8/grunt-cssbeautifier).

The full CSS file that was used is:

Minified via ``cssmin`` (causing the problem): https://dev.dermarolling.nl/wp-content/themes/blanco/css/full.raw.css
Beautified: https://dev.dermarolling.nl/wp-content/themes/blanco/css/full.css

The used penthouse conifguration is the following:

```json
penthouse: {
    frontpage : {
        outfile : 'css/abovethefold/frontpage.css',
        css : 'css/full.css',
        url : 'https://dev.dermarolling.nl/',
        width : 1600,
        height : 1200
    },
    product: {
        outfile : 'css/abovethefold/product.css',
        css : 'css/full.css',
        url : 'https://dev.dermarolling.nl/product/titanium-dermaroller/',
[...]

PR #38, opened 2015-01-17
fix for bug introduced in 0.2.52

pocketjoso:
Solves #36 and #37. Whitespace issues between my editors make for a noisy commit.

* `splitsWithIndexes` was not thought through properly, causing major bugs in critical path css generation, and is therefor reverted.
* Issue #36 now instead properly solved; the issue was missing to step `selectorsKept` when `::`selectors are encountered (as they are not tested but just simply kept in critical css).
* Added a more realistic critical path css generation test, taken from [critical's tests](https://github.com/addyosmani/critical/blob/master/test/test.js) - should prevent future bugs from getting merged in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,12 +27,12 @@ Install [PhantomJS](https://github.com/ariya/phantomjs) first, and make sure it 
 #### Usage
 
 	phantomjs penthouse.js [URL to page] [CSS file] > [critical path CSS file]
-	
+
 	//for example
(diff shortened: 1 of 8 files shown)


pocketjoso:
As [Critical](https://github.com/addyosmani/critical)'s test picked up the issued `Penthouse's` test didn't, I think I'll just grab a test from there, for now.

PR #44, opened 2015-02-08
fix: options parser fixes

pocketjoso:
The options-parser was erroring on relative html paths, but the error it throws was not passed on by `core.js`, unless in `debug` mode. Instead the feedback was just the usage string, which was very confusing. In current master such relative html paths were just let through and worked fine. (I was never aware of this until now though).

To conclude: If the options parser is going to stop execution on errors that can't be easily understood via the usage string produced by `core.js`, then the logic needs to be changed so that core displays those errors. For now, as the invalid url guard disappears, no change in core is needed.

### DIFF
--- a/lib/options-parser.js
+++ b/lib/options-parser.js
@@ -1,12 +1,13 @@
 /*
-parser for the script - can be used both for the standalone node binary and the phantomjs script
-*/
+ * parser for the script - can be used both for the standalone node binary and the phantomjs script
+ */
+
 /*jshint unused:false*/
 
 var usageString = '[--width <width>] [--height <height>] <url> <main.css>';
 
 function buildError(msg, problemToken, args) {
-    var error = new Error( msg  + problemToken);
+    var error = new Error(msg + problemToken);
     error.token = problemToken;
     error.args = args;
     throw error;
@@ -17,25 +18,25 @@ function buildError(msg, problemToken, args) {
 // throws an error on wrong options or parsing error
 function parseOptions(argsOriginal) {
     var args = argsOriginal.slice(0),
-    validOptions = ['--width', '--height'],
-    parsed = {},
-    val,
-    len = args.length,
-    optIndex,
-    option;
+        validOptions = ['--width', '--height'],
+        parsed = {},
+        val,
+        len = args.length,
+        optIndex,
+        option;
 
-    if(len < 2 ) buildError('Invalid number of arguments', args, args);
+    if (len < 2) buildError('Invalid number of arguments', args, args);
 
-    while(args.length > 2 && args[0].match(/^(--width|--height)$/)) {
+    while (args.length > 2 && args[0].match(/^(--width|--height)$/)) {
         optIndex = validOptions.indexOf(args[0]);
-        if(optIndex === -1) buildError('Logic/Parsing error ', args[0], args);
+        if (optIndex === -1) buildError('Logic/Parsing error ', args[0], args);
 
         // lose the dashes
         option = validOptions[optIndex].slice(2);
         val = args[1];
 
         parsed[option] = parseInt(val, 10);
-        if(isNaN(parsed[option])) buildError('Parsing error when parsing ', val, args);
+        if (isNaN(parsed[option])) buildError('Parsing error when parsing ', val, args);
 
         // remove the two parsed arguments from the list
         args = args.slice(2);
@@ -44,16 +45,21 @@ function parseOptions(argsOriginal) {
     parsed.url = args[0];
     parsed.css = args[1];
 
-    if( ! parsed.url.match(/https?:\/\//) ) {
-      buildError('Invalid url: ', parsed.url, args);
+    if (!parsed.url) {
+        buildError('Missing url/path to html file', '', args);
     }
+
+    if (!parsed.css) {
+        buildError('Missing css file', '', args);
+    }
+
 
     return parsed;
 }
 
-if(typeof module !== 'undefined') {
+if (typeof module !== 'undefined') {
     module.exports = exports = {
-        parse : parseOptions,
-        usage : usageString
+        parse: parseOptions,
+        usage: usageString
     };
 }
--- a/lib/options-parser.js
+++ b/lib/options-parser.js
@@ -25,7 +25,7 @@ function parseOptions(argsOriginal) {
         optIndex,
         option;
 
-    if (len < 2) buildError('Invalid number of arguments', args, args);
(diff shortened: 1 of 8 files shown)


lib/options-parser.js
```diff
@@ -44,6 +46,7 @@ function parseOptions(argsOriginal) {
     parsed.url = args[0];
     parsed.css = args[1];
 
+    // this is problematic...
```
pocketjoso: The below guard fails the tests for [critical](https://github.com/addyosmani/critical/blob/master/test/test.js#L50), as they're passing relive paths to `Penthouse`, f.e. `test/fixture/index.html`.

I just looked quickly and couldn't see how they get away with doing that, as phantomjs will always fail to connect to URL's when I leave out the `protocol`...

Since it must be possible to do so in some way, other people might be doing it as well, and I therefor think we have to drop the check below completely.. If it errors, it will be caught inside the core script anyway.

pocketjoso:
@fatso83 I just need to resolve these two issues, then I'm happy to merge to master. Any input? Especially on how to properly surface the errors thrown in buildError, the `console.log` I added is very hacky.

fatso83:
Looking at it now.

lib/options-parser.js
fatso83: That seems way premature. Don't remove things if you don't know IF they actually pose a problem, as this test intend to serve the purpose of early warning :-) As `critical` is just using the file paths, they must themselves be using some kind of server (like we are doing in the tests) to serve the files. Not rocket science :-)

```
var app = module.exports = express();
app.use(express.static(__dirname ));
app.listen(port)
```

fatso83: You are right - we do need to remove the check ... 

```
$ phantomjs penthouse.js ~/dev_priv/penthouse/test/static-server/page1.html  ~/dev_priv/penthouse/test/static-server/page1.css 
#box1 {
    width : 800px;
    height: 400px;
    background : red;
}
#box2 {
    width : 800px;
    height: 400px;
    background : blue;
}

#box21 {
    width : 100%;
    height: 100px;
    background: yellow;
}

#box22 {
    width : 100%;
    height: 50px;
    padding-top: 200px;
    background: green;
}
```
There is no need to fire up a server, local file paths work just fine.

fatso83:
Not completely finished reviewing everything. I am pushing some changes to my own repo, but one small tip: if you look into `make.js` you will see several targets defined (test, lint, all, ...). Try running `./make.js lint` at the end of a big Pull Request and it will find lots of potential problems for you. I have found lots of unused code/variables so far.

```
$ ./make.js lint
lib/phantomjs/core.js: line 353, col 34, 'options' was used before it was defined.
lib/phantomjs/core.js: line 41, col 28, 'cssPreformatter' is not defined.
lib/phantomjs/core.js: line 75, col 21, 'unusedFontfaceRemover' is not defined.
lib/phantomjs/core.js: line 358, col 13, 'parseOptions' is not defined.
lib/phantomjs/core.js: line 359, col 13, 'usageString' is not defined.
lib/phantomjs/core.js: line 28, col 35, 'trace' is defined but never used.
lib/phantomjs/core.js: line 28, col 28, 'msg' is defined but never used.

lib/phantomjs/css-preformatter.js: line 44, col 2, Unnecessary semicolon.

lib/phantomjs/unused-fontface-remover.js: line 60, col 2, Unnecessary semicolon.

9 errors
```

fatso83:
I have made some corresponding changes in a branch with the same name in my fork. You could pull those. Should answer your two issues :)

lib/options-parser.js
pocketjoso: Hmm, so down the line we can simplify the tests too - no need for us either to start up a server.

Will remove this check now.

pocketjoso:
> How have you tried testing this? The tests for this module clearly shows that the error is thrown and propagated to the calling function. See parser-test.js

Only so far as running Penthouse through the CLI and through the Node module, and seeing that the `buildError` message get's lost. I also saw that it works as it should in the parser test, as you say.

pocketjoso:
> ...if you look into make.js you will see...

Yeah, I'm familiar with that code now. I did fix all errors in `core.js`, except for ones that are tricky due to how we concatenate files. I didn't see any lint errors for any other files though.

pocketjoso:
> I have made some corresponding changes in a branch with the same name in my fork...

In your fork, in the branch `options-parser-fixes` I see my own latest commit as the latest one - nothing added from you?

pocketjoso:
@fatso83 I can't see changes in your fork, but I just looked more closely in to how the error got passed along in the `core.js`, and unless you're in `debug` mode, it's just ignored. Hence the silence. I might change that..

pocketjoso:
To conclude: If the options parser is going to stop execution on errors that can't be easily understood via the usage string produced by `core.js`, then the logic needs to be changed so that core displays those errors. For now, as the invalid url guard disappears, no change in core is needed.

fatso83:
Sorry, I closed my laptop before I saw it fail. Tried pushing to your repo, not mine. Pushed now.

I think the reason I did not understand what you were talking about, concerning errors not showing, was due to me using the Node standalone, which does parsing before calling the `penthouse()` function, instead of waiting for the PhantomJS process to fire up and exit on error.

```
$ bin/penthouse --width asd   foobar.css
Caught error parsing arguments:  Parsing error when parsing asd

penthouse [--width <width>] [--height <height>] <url> <main.css>
```

After running the standalone, I now understood what you were saying. Still, the last commit (9fce9eb) is not a solution. That should be reverted, as all that does is suppress throwing an error, not fixing the actual logging issue. Other parsing errors would still not show!

I have now pushed some additional commits fixing this, as well as the previous issues. It now checks to see if the script is running in standalone mode and outputs additional error messages if that is the case.

pocketjoso:
Okay, got your commits in now. Dropped my ones, as they weren't contributing to the final solution.

pocketjoso:
Errors returned are not quite right..

```
$ ./bin/penthouse test/static-server/yeoman.html
Caught error parsing arguments:  Invalid number of argumentstest/static-server/yeoman.html

penthouse [--width <width>] [--height <height>] <url> <main.css>
```
If you add in some option, then it works:
$ ./bin/penthouse --width 500 test/static-server/yeoman.html
Caught error parsing arguments:  Missing css file

penthouse [--width <width>] [--height <height>] <url> <main.css>
```

pocketjoso:
Okay, this looks good enough for now, will merge.

PR #46, opened 2015-02-12
0.3.0

pocketjoso:
This is going to be a big merge.
I intend to cleanup the history here, and drop all commits that didn't end up in `0.3.0`. Eventually I will replace the current `master` with this branch, as they diverged a long time ago. I'm going to look into [this merge strategy](http://stackoverflow.com/a/2763118), so we keep the history.


pocketjoso:
Undertaking the commit history cleanup in branch `0.3.0-cleanup..`

PR #47, opened 2015-02-12
0.3.0 cleanup

pocketjoso:
Cleaning up for 0.3.0 merge..


pocketjoso:
Rebased (mutliple times), now managed to mostly remove the trace of `multi url` handling through these commits. Am now going to remove the need for running a server in `tests`, then merge.

PR #55, opened 2015-05-31
add version argument to CLI

pocketjoso:
Want to be able to handle `penthouse -v`, not sure this is the right way. cc @fatso83 .

### DIFF
--- a/bin/penthouse
+++ b/bin/penthouse
@@ -6,9 +6,15 @@
  */
 
 var penthouse = require('../lib/index.js'),
+    version = require('../package.json').version,
     parser = require('../lib/options-parser.js'),
     args = process.argv.slice(2),
     options;
+
+if(args.length === 1 && args[0] === '-v' || args[0] === '--version'){
+  console.log(version);
+  process.exit(0);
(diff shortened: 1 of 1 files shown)


fatso83:
Seems fine. But wouldn't it be more appropriate to put it in the module actually handling all the other options parsing? Single Responsibility Principle, etc.

pocketjoso:
Thought it would be awkward because still have to do the check outside, but updated now.

fatso83:
You know, I am an idiot at times... Your first attempt was the best, since
the last commit will produce failing code AFAI can tell. The require call
(in options-parser) will work in Node-code, but will fail to resolve at
run-time for PhantomJS. If you need a way of querying the version number
for the phantomjs script (without Node), the version number will have to be
built in at build time. That seems awkward ... So I guess it's mostly just
relevant for Node, in which case your first shot was probably the best.

Sorry for the slow thinking.

pocketjoso:
That's all right, was kind of what I intuitively thought. Will just drop the fixup and merge.

Cheers.

PR #57, opened 2015-06-03
Handle html string

pocketjoso:
Add support to pass in a `html` string instead of `url`. Should perhaps go as separate option for node module. Fine to not support at all for CLI.

Cannot figure out why `page.content = <html string>` produces ever so slightly different result than `page.open(<html url>)` in `PhantomJS`, failing the test I added... Html is the same, viewport dimensions are the same, but critical css comes back with some extra or missing rules, indicating that something is not taking up the same space on the page (or the the viewport dimensions are incorrect)...

### DIFF
--- a/lib/phantomjs/core.js
+++ b/lib/phantomjs/core.js
@@ -107,6 +107,228 @@ page.onCallback = function(css) {
   }
 };
 
+var evalutePageCss = function(options) {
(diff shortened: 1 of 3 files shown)


pocketjoso:
Going to park this for now, wait for `phantomjs 2`.

Issue #58, opened 2015-07-07
Suppress errors from phantomjs?


bitinn:
I sometimes see these lines in the output css when running penthouse, is there any way to prevent phantomjs from messing with output?

```
Unsafe JavaScript attempt to access frame with URL about:blank from frame with URL file:///Users/local/git/mai/node_modules/penthouse/lib/phantomjs/core.js. Domains, protocols and ports must match.
```

my command look something like this:

```
penthouse https://local.dev/ public/app.css --width 1366 --height 768 > public/inline.css
```

fatso83:
Can you make this into a reproducible case? I would like to look at it when given the time, but make it easy for me and it will be sooner rather than later ;-)

- steps taken to reproduce
- operating system
- version of penthouse
- how was it installed?
- expected output
- actual output (given above), but supply the actual full output

bitinn:
I am ok with investigating this myself, I am following leads on this 2 threads:

https://github.com/ariya/phantomjs/issues/12697
https://github.com/n1k0/casperjs/issues/1068

I believe either:

1. adding `webSecurityEnabled: false` to your phantomjs config.json (I tried and it works). [ref](http://phantomjs.org/api/command-line.html)
2. wrap all `phantom.exit` in setTimeout to discard these stdout warnings. [ref](https://github.com/jonkemp/qunit-phantomjs-runner/commit/afe6f276024ab86aa2dbb649263ec8840e001187)
3. the cause for this error is due to phantoms 1.9.x changing url on exit and triggers CSP, this issue does not exist on 2.x, so upgrading penthouse support to phantomjs 2.x also works.

From this PR discussion I believe the issue won't be fixed on 1.9.x branch

https://github.com/ariya/phantomjs/pull/12720

Solution for fix 1 and 2 are relatively simple, if you prefer I can make a PR

bitinn:
PS: yet another alternative is to downgrade to `phantomjs: 1.9.11` which is the last wrapper to use phantomjs 1.9.7, which doesn't have this problem.

Which direction would you prefer?

pocketjoso:
@bitinn From my quick readup on this problem upgrading `phantomjs` to `2.x` should solve the problem. The only reason I haven't is because of lack of official Linux release, so if you're on a non Linux platform, I would suggest upgrading.

As for a fix for this release, I should be able to supress the error from within `Penthouse`. If you solve it before me, good, otherwise will look into this weekend.

Cheers

[...]

PR #60, opened 2015-07-08
Wrap phantom exit in setTimeout

bitinn:
This PR implements the solution 2 in #58 

Pro: will work with phantomjs core v1.9.8, so it works with latest nodejs wrapper (currently at v1.9.17).

Con: hopefully no one will need to debug phantom exit output.

It works consistently on my repo, let's see if I broke any travis tests...

### DIFF
--- a/lib/phantomjs/core.js
+++ b/lib/phantomjs/core.js
@@ -21,6 +21,16 @@ var debug = function() {
 	if (DEBUG) errorlog('DEBUG: ' + combineArgsString(arguments));
 };
 
+// discard stdout from phantom exit;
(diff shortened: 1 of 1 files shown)


pocketjoso:
Hi, thanks for the pr! There might be a cleaner way to do this; I will have a quick look this weekend, and if I don't see a better solution I will merge your PR.

pocketjoso:
I'm going to merge this as a fix for this problem. It's a bit ugly and hopefully we can just remove this code when we upgrade to `phantomjs 2.x`.

Issue #61, opened 2015-07-10
Force include


bitinn:
Not a big deal, but nice to have:

Imagine you have a header at the top of page that shows currently login user (otherwise display a login button), because such info is only visible to login users, `penthouse` won't be able to see that this class is used (it will only inline the login button's css).

While one can argue most existing users should have a cached copy of full css in their browser anyway, there are indeed cases where a login user can run into problems (though eventually the full css will load and all should be fine at that point).

This does make `penthouse` look a bit like [criticalCSS](https://github.com/filamentgroup/criticalCSS), which has this feature but lack self-signed ssl support for local development. But I guess these 2 tools serves very similar purpose so it's natural to see their feature set converges :)

pocketjoso:
Hi, thanks for opening this issue. While critical css is most important for new users, I do agree that it's also useful for logged in users, as they can remain logged while their cached css has become redundant due to new release.

I'm not sure I'm in favour of `forceInclude`. It moves the critical css task away from being automated, which is what I really strive for it to be. It also ties you to your current css and html/js, meaning you have to keep updating it all the time. The manual work going in to figuring out what selectors need to be forced can also already today be used into just concatenating the `critical css` given by `Penthouse` with the extra css you've figured out you need for `logged in` users.

I'm thinking an ability to set a `cookie` with login info would be more useful; you can then produce a logged in and a logged out critical css for each page. Would that solve your use case?

And yeah of course `Penthouse` looks similar to `criticalCSS`; these two tools basically do the same thing.

bitinn:
> I'm thinking an ability to set a cookie with login info would be more useful

I agree, something like a `--cookie` flag should work and keep the burden away from developers. But I have no experience with phantomjs so can't tell if it's possible.

pocketjoso:
You can set cookies with `phantomjs`, but that approach won't work for session cookies... I'll have a think about it.

pocketjoso:
Setting a `session cookie` with phantomJS in `Penthouse` requires too much custom logic to be worthwhile.

@bitinn Would it be possible for you to generate html for logged in pages in your build process, and pass these files to `Penthouse`? That would bypass the need for `Penthouse` to deal with any auth logic - it would just generate `critical css` for the `html` provided.

bitinn:
> Would it be possible for you to generate html for logged in pages in your build process, and pass these files to Penthouse?

[...]

PR #64, opened 2015-07-18
Force include 

pocketjoso:
For #61.

Node module now takes an array of selectors that will not be removed even if they don't appear in the critical viewport:
``` 
penthouse({
  ...
  forceInclude: [ '.myLoggedInSelector', '.andFriends', /^\.andRegexToo.*/ ] 
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,10 @@ penthouse({
     // OPTIONAL params
     width : 1300,   // viewport width
     height : 900,   // viewport height
+    forceInclude : [
+      '.keepMeEvenIfNotSeenInDom',
+      /^\.regexWorksToo/
+    ],
     timeout: 30000 // ms; abort critical css generation after this timeout
 }, function(err, criticalCss) {
     if (err) { // handle error }
--- a/lib/index.js
+++ b/lib/index.js
@@ -26,13 +26,23 @@ var m = module.exports = function (options, callback) {
         debuggingHelp = '',
         cp
 
-    // set the options, falling back to defaults
-    var scriptArgs = [
-      options.url || '',
-      options.css || '',
-      options.width || DEFAULT_VIEWPORT_WIDTH,
-      options.height || DEFAULT_VIEWPORT_HEIGHT
-    ]
+  // need to annotate forceInclude values to allow RegExp to pass through JSON serialization
+  // also need to give them a toJSON method
+  RegExp.prototype.toJSON = RegExp.prototype.toString;
+  var forceInclude = (options.forceInclude || []).map(function (forceIncludeValue) {
+    if (typeof forceIncludeValue === 'object' && forceIncludeValue.constructor.name === 'RegExp') {
+      return { type: 'RegExp', value: forceIncludeValue.source };
+    }
+    return { value: forceIncludeValue };
+  })
+  // set the options, falling back to defaults
+  var scriptArgs = [
+    options.url || '',
+    options.css || '',
+    options.width || DEFAULT_VIEWPORT_WIDTH,
+    options.height || DEFAULT_VIEWPORT_HEIGHT,
+    JSON.stringify(forceInclude) // stringify to maintain array
+  ]
 
   cp = spawn(phantomJsBinPath, [configString, script].concat(scriptArgs));
 
--- a/lib/phantomjs/core.js
+++ b/lib/phantomjs/core.js
@@ -128,7 +128,7 @@ function getCriticalPathCss(options) {
 			// sandboxed environments - no outside references
 			// arguments and return value must be primitives
 			// @see http://phantomjs.org/api/webpage/method/evaluate.html
-			page.evaluate(function sandboxed(ast) {
+			page.evaluate(function sandboxed(ast, forceInclude) {
 				var h = window.innerHeight,
 					renderWaitTime = 100; //ms TODO: user specifiable through options object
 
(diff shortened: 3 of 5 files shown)


pocketjoso:
@fatso83 If you have time please have a look at the options parsers commit (https://github.com/pocketjoso/penthouse/commit/609ff041a23ac7c207ae2259d6b1da8f43a7fc1a) - what's the proper way of handling different option types? hard code the order?

pocketjoso:
@bitinn Do you want to play around with this before I merge? Should do what you wanted.

bitinn:
Yep, tested, it appears to work as expected.

bitinn:
Oh I should add that `--forceInclude` only works when coming after `penthouse`, so it does not work when I write `penthouse [url] [css] --forceInclude`

Might be easier to use something like https://github.com/substack/minimist

pocketjoso:
> Oh I should add that --forceInclude only works when coming after penthouse, so it does not work when I write penthouse [url] [css] --forceInclude

Yeah I know..

> why roll your own?

Yeah i've been wondering about this too. @fatso83 would you agree it would be easier to use a library for the options parsing, or anything I'm missing here?

fatso83:
that hack is proper ugly, indeed 😷

I am delivering a thesis Wednesday, so I won't have time to look at the specifics of this until later, but I thought I might clear up to @bitinn why we actually needed to roll our own parsing. I am not affected by the Not Invented Here syndrome, and prefer to use something premade when _possible_.

The main reason is that this is not a Node project, but mainly a PhantomJS project and needs to work without Node. @pocketjoso has been clear from the start that he wants this project to be useful by people that have not installed Node, so any option parsing has to work within the phantomjs script. As such we cannot use node modules directly. The current options parser is shared by both the Node wrapper and the standalone PhantomJS script.

Of course, there is nothing stopping us from ripping the insides of any node parsing library and modifying it to work with PhantomJS, but as we for the last two years have only needed to support _two_ options, dashing out 15 lines of code was a far easier option ;-)

Now that there are new requirements it might be worth investing time in adapting an existing library, but it is not necessarily the best approach if the number of options we need to support is minimal. Whatever we end up with, the suggested [hack](https://github.com/pocketjoso/penthouse/commit/609ff041a23ac7c207ae2259d6b1da8f43a7fc1a)  is definitively not something that can remain. Way too fragile. But maybe it can stay in the dev branch for this feature until we have cooked up something better?

Requirements for any new options parsing:
- no requirements on any other Node libraries
- it must continue working with the standalone PhantomJS script (`phantomjs penthouse.js --arg1 --arg2 ...`)
- still work with the [build process](https://github.com/pocketjoso/penthouse/blob/master/make.js#L33) (concatenated into one bigger script)
- pass all current tests of course

pocketjoso:
Thanks for chiming in @fatso83 - do focus on your thesis! I don't think we need to rush this (correct me if you're in a hurry to have this functionality @bitinn). That commit wasn't meant to be kept around, just there as a proof of concept to replace. :)

I wonder about my initial assumption that some people will use CLI without `Node`; sounds like an exotic combo to me now.. I'll have a think about that.

bitinn:
> it must continue working with the standalone PhantomJS script

In this case, I suggest a longterm plan to migrate to PhantomJS 2, there are likely new features in v2 that will make this library easier to implement.

Otherwise, go for a new major version of `penthouse` that depends on other npm modules, it will likely make `penthouse` core leaner and more maintainable.

@pocketjoso don't worry about my use case for now, I have changed `penthouse` viewport size and my css classes to workaround my specific problem in #61. Thanks for following up.

fatso83:
@pocketjoso dropping the requirement that everything must work the same in the standalone PhantomJS version would make a lot of things potentially easier. one thing is this silly options parsing thing, but there are other things that was too cumbersome to implement given that requirement, like spinning up multiple instances controlled by one node process (that we talked about at one point).

bitinn:
Just discover another interesting use-case of force include: if you have an class that's applied before your main css/js files are loaded, say a `loading` class, and it's meant to be removed when dom is ready (or onload), `penthouse` won't see it in DOM and hence output doesn't contain such class (which more or less defeat it's purpose).

Maybe force include is not necessary to achieve this, but I can't think of a way to trick it into `penthouse` output just yet...

pocketjoso:
@bitinn Yeah that's completely true. It's problematic for `Penthouse` to capture that such a class should be included (without `forceInclude`).

pocketjoso:
Heads up @bitinn - now merging this. Added `regex` support, will publish new `npm` version in a few mins. :)

clanwce:
I tried this and it doesn't seem to work. I included some extra classes that might be on the page such as the classes for a featured post that only shows sometimes. The css I passed to penthouse contains the styles for these classes but they are not in the final generated critical css.

pocketjoso:
@clanwce The only reason I can think of for this to happen would be if your original css contains errors that confuse the AST parser. Try running penthouse with `strict:true` and see if it complains about such errors.

clanwce:
@pocketjoso Thank you for replying so soon. I tried setting strict to true and run the script, it didn't complain about any error, but it didn't extract css for the classes I put into forceInclude either.

clanwce:
I'd like to know how force include work. Does it work when the classes you put in it doesn't exist in the html page? My assumption is that it will check the classes in forceInclude array and extract all related css from the css file that's passed to penthouse and it doesn't matter if those classes exist in html or not. If that's not the case, then that could be why it doesn't work because the classes I put there may not be in the html file.

pocketjoso:
@clanwce 

> I'd like to know how force include work

You're commenting on the PR that implements it, but yes, you are right. It just ensures that any selectors matching the `forceInclude` list, found in the css you pass in, will remain in the critical css - regardless of whether they are found on the page or not.

Your problem should be related to your original css. Could you send me a public url to test towards, either here, or via f.e. twitter DM (my handle is the same). I'll check it out for you.

Cheers

PR #59, opened 2015-07-08
Disable web-security

bitinn:
To suppress phantom exit error, as describe in https://github.com/pocketjoso/penthouse/issues/58

Pro: simple and straight-forward.

Con: potential security issue if you build in production environment.

PR #65, opened 2015-07-19
lower phantomjs security settings

pocketjoso:
Ignoring more errors thrown by `phantomjs`, for #58 .

Untested.

### DIFF
--- a/lib/phantomjs/core.js
+++ b/lib/phantomjs/core.js
@@ -31,8 +31,12 @@ var phantomExit = function(code) {
 	}, 0);
 };
 
-//don't confuse analytics more than necessary when visiting websites
+// don't confuse analytics more than necessary when visiting websites
 page.settings.userAgent = 'Penthouse Critical Path CSS Generator';
+
+// we don't really care about web security for generating the css
+page.settings.webSecurityEnabled = false;
+page.settings.localToRemoteUrlAccessEnabled = true;
 
 /* prevent page JS errors from being output to final CSS */
 page.onError = function(msg, trace) {


pocketjoso:
@bitinn Can you test whether this (and the previous fix) resolves your security error? Thanks. :)

bitinn:
I don't think `webSecurityEnabled` will be effective in resolving my specific issue, as mentioned in #59.

Not sure about `localToRemoteUrlAccessEnabled`, from the doc it seems to apply to local file? And these 2 settings do impact security.

pocketjoso:
Thanks @bitinn. That fix didn't solve [everyone's problems](https://github.com/addyosmani/critical/issues/92#issuecomment-122633791), hence this PR.

Not sure what the best way to solve this is, as it's all temporary until we get an official `phantomjs 2` release for `linux`. I see that `criticalCSS` just [replace these errors after phantomjs execution](https://github.com/filamentgroup/criticalCSS/blob/master/critical.js#L61-L64).. doesn't seem to be a clean way, maybe that's as good as it gets.

Will probably release this as a `release candidate` and ask if it solves the problems, as I can't reproduce them consistently.

bitinn:
Guess there is no better way to resolve this: you either ignore those errors using setTimeout or handle them during stdout.

Could you let the developer seeing this problem to change your config.json manually and see if that issue is resolved. Also `penthouse` is using that phantom config right, hopefully it's not a legacy config as well?

ref: https://github.com/pocketjoso/penthouse/pull/59/files

pocketjoso:
> Also penthouse is using that phantom config right?

`Penthouse` does use the `lib/phantomjs/config.json` file: https://github.com/pocketjoso/penthouse/blob/master/lib/index.js#L25, so you're right, if anything this PR should update the config rather than modifying the script.

> Could you let the developer seeing this problem to change config.json manually and see if that issue is resolved

I'll reach out to other guy with this problem and see if he can try to update the config to see if that resolves the problem.

pocketjoso:
[Wasn't enough](https://github.com/addyosmani/critical/issues/92#issuecomment-122666396), replaced by #66.

PR #66, opened 2015-07-22
silence known phantomjs stdout security errors

### LINKED ISSUE #58: Suppress errors from phantomjs?

pocketjoso:
Fix for known phantomjs security errors going to `stdout`, resolves #58 and related issues.

Not the best way to solve this, but again these problems should disappear in `phantomjs 2`. Solution pretty much copied from [criticalCSS](https://github.com/filamentgroup/criticalCSS/blob/master/critical.js#L61-L64).

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -9,6 +9,13 @@ var phantomJsBinPath = phantomjs.path;
 var configString = ('--config=' + path.join(__dirname, 'phantomjs', 'config.json'));
 var script = path.join(__dirname, 'phantomjs', 'core.js');
 
+var removePhantomJSSecurityErrors = function (stdOut){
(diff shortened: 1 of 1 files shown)


pocketjoso:
Note to self: write test to ensure returned css is valid!

Issue #71, opened 2015-08-19
Timeout penthouse after x seconds


pocketjoso:
Under some circumstances, `PhantomJS` just hangs (as an example: #70), and at other times response times can go well over 1 min.

I think by default Penthouse should return an error after `30s`, but this timeout value should be configurable as an option.

pocketjoso:
Resolved by #73.

PR #73, opened 2015-08-20
Timeout

pocketjoso:
For #71.

Requires #81.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,10 @@ var penthouse = require('penthouse'),
 penthouse({
     url : 'http://google.com',
     css : path.join(__basedir + 'static/main.css'),
+    // OPTIONAL params
     width : 1300,   // viewport width
-    height : 900   // viewport height
+    height : 900,   // viewport height
+    timeout: 30000 // ms; abort critical css generation after this timeout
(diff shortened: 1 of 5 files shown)


pocketjoso:
Okay, do I really have to send exit code `0` in `process.exit` handler in order to not fail tests? Oo Doesn't seem right to me..

fatso83:
> do I really have to send exit code 0 in process.exit handler in order to not fail tests

No, you are already in the exit signal handler, and so you should not repeat the call. That might incur weird stuff, like you experience.

pocketjoso:
Cool, I will merge once we update the options parsing like I said; don't want to merge it before as people could potentially be relying on runs with really long execution times (that would from hereon throw errors).

pocketjoso:
Added a commit using `phantomjs'` `resourceTimeout` flag, have done some quick tests and it also seems to help a lot for the issues with pages otherwise just timing.

Putting it in the same PR as they really solve the same problem - and both these options need to be configurable in the node module, so both need to wait for overhaul of options parsing.

pocketjoso:
Rebased on AST `penthouse` and now used as an option - ready to go!

PR #78, opened 2015-10-18
Ast in core

pocketjoso:
Delegates all CSS parsing to the `css` parsing node module, allowing the `Penthouse` core to work on an AST object instead. Reduces lines of code, and should reduce bugs.

Should break the standalone `penthouse.js` file, so need to update README to reflect this.

I'm going to give this some solid testing (on top of the unit tests; which do succeed) before I merge.


pocketjoso:
cc @fatso83 - now removing command line support. Will merge tomorrow.

PR #81, opened 2015-11-01
Drop custom options parser

pocketjoso:
No longer needed as standalone Penthouse support has been dropped. cc @fatso83


pocketjoso:
I realise that really there's no need to drop the binary here, instead it could be changed to make use of a node options parser like `commander` or `nopt`, but that requires separate work from what I'm looking at now. As I don't want to have to update the custom options-parser that I'm now removing as I add in more options (`timeout`) I will remove the binary for now, but it should be brought back with updated options parsing in the future (at least if anyone cares for it).

pocketjoso:
Not necessary to remove binary - all that should happen is for it to make use of a npm package for options parsing, such as `commander`. I had a quick go at it now but couldn't get it to play ball, so just removing for now.

PR #88, opened 2015-11-15
Remove unmatched media queries

### LINKED ISSUE #67: Strip out unused media queries
Larger than the defined viewport.

pocketjoso:
`print` media queries were already removed, now `Penthouse` also removes and `min-width` / `min-height` media query that exceeds the defined viewport size (1300x900 by default).

Resolves #67.

### DIFF
--- a/lib/phantomjs/core.js
+++ b/lib/phantomjs/core.js
@@ -2,8 +2,9 @@
 
 var cssAstFormatter = require('css'),
   embeddedbase64Remover = require('./embedded-base64-remover'),
-  ffRemover = require('./unused-fontface-remover.js'),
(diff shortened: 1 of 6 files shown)


josenobile:
The unmatched min-width should be restored. We don't which device will receive the content. See #145

pocketjoso:
Penthouse only removes media queries that **exceeds** the specified viewport dimensions, i.e. all media queries for smaller viewports will still remain. This means that a mobile first design will be kept intact in the critical css, the only thing that is dropped by default is media queries for very large screens; and this can be modified by passing in `width` and `height` parameters.

PR #100, opened 2015-12-28
Remove unused CSS properties

bevacqua:
Added `apartment` to remove `animation` and `transition`, including all vendor prefixes and destructured properties.

I was wondering if we could benefit from adding an option that removes a few more properties that don't really affect layouting and could further reduce byte size:

- `::selection`
- `text-shadow`
- `box-shadow`
- `border-radius` properties with `px` values under `50px`
- `pointer-events`
- `cursor` if we want to be super aggressive
- I'm sure there's others

### DIFF
--- a/lib/phantomjs/core.js
+++ b/lib/phantomjs/core.js
@@ -3,6 +3,7 @@
 var cssAstFormatter = require('css'),
   embeddedbase64Remover = require('./embedded-base64-remover'),
   ffRemover = require('./unused-fontface-remover'),
+  apartment = require('apartment'),
   fs = require('fs'),
   nonMatchingMediaQueryRemover = require('./non-matching-media-query-remover'),
   system = require('system'),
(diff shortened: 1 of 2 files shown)


pocketjoso:
Thanks for the PR, it's awesome.

I've always drawn the line for what to include in criticalcss at making sure the (critical part of the) page still looks the same, so following that I think you can safely add in:
* `pointer-events`
* `cursor`

The other properties you suggest do effect the first render and I think should therefor not be part of the default config. Perhaps we could provide an additional more aggressive config, or just document more aggressive usage in README (which really needs a general overhaul - I'm behind on that).

`::selection` is a selector and not a property - does apartment really handle that? If it does it fits my criteria for what to remove so I would be for adding that in to the PR as well.

If you can make does small changes to you PR I'll merge it afterwards!

Cheers

bevacqua:
There we go. Updated `apartment` to deal with selectors too

pocketjoso:
Okay, I'll merge that for now, although as I said I'm already checking the selectors in the main loop of the `Penthouse` core, so eventually I will move the `::selection` selector there and then no longer use `apartments` selector handling.

PR #103, opened 2016-01-16
fix memory leaks

pocketjoso:
process event listeners were never cleared explicitly
(only by the parent process exiting), which meant that in long living
node processes all variables in scope were retained forever.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -26,8 +26,10 @@ var m = module.exports = function (options, callback) {
   var stdOut = '',
     stdErr = '',
     debuggingHelp = '',
-    cp
+    cp,
+    killTimeout
 
+  var timeoutWait = options.timeout || DEFAULT_TIMEOUT
   // need to annotate forceInclude values to allow RegExp to pass through JSON serialization
   var forceInclude = (options.forceInclude || []).map(function (forceIncludeValue) {
     if (typeof forceIncludeValue === 'object' && forceIncludeValue.constructor.name === 'RegExp') {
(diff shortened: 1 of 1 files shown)


pocketjoso:
I'm not an expert on Node memory management, this has been my first deep dive into it, so I'm hope my conclusion is correct. What brought me here was that my node server serving critical css using penthouse for https://jonassebastianohlsson.com/criticalpathcssgenerator/ was using more and more memory over time, and I eventually tracked it down to these process event listeners.

Any feedback? @fatso83 @bezoerb 

Will test on my own servers before I merge.
Cheers

pocketjoso:
The files changed diff is a bit confusing due to that I moved around some other code as well. The relevant lines for this PR are:
https://github.com/pocketjoso/penthouse/blob/fix-memory-leaks/lib/index.js#L120-L122

pocketjoso:
Works well in my tests so will merge - still interested in feedback on whether my changes seem reasonable to someone with more node memory management experience, so feel free to comment.

lib/index.js
```diff
@@ -115,14 +113,22 @@ var m = module.exports = function (options, callback) {
...
       callback(err)
     }
-  })
-
-  process.on('exit', function () {
+    // we're done here - clean up
+    clearTimeout(killTimeout)
+    // can't rely on that the parent process will be terminated any time soon,
+    // need to rm listeners and kill child process manually
+    process.removeListener('exit', exitHandler)
```
fatso83: Yup, this seems needed :-)

fatso83:
Just some general feedback on memory profiling and finding memory leaks. To be sure that you remove the right things there are some things that will make it easier:
- a way of reproducing the memory leak
- heap snapshots 

Force GC if possible.
Use the Node inspector tool and take a snapshot.
Then do some things which will make the memory usage increase.
Force GC if possible.
Take a new snapshot.

Then use DevTools in Chrome to load and diff the snapshots. That will show you which objects increased in numbers, and you can probably find out from where. [This Chrome tutorial](https://developer.chrome.com/devtools/docs/javascript-memory-profiling) shows how it is done.

pocketjoso:
Cheers @fatso83. I read all these tutorials but it was still a bit confusing to me initially. In the end I used [heapdump](https://github.com/bnoordhuis/node-heapdump), which forces gc and then saves heapdumps, and I "called it" during runtime via running `kill -USR2 <node process pid>` in shell. I then used the `comparison` view in the `profiler` in Chrome to see what had accumulated between two heapdumps, and finally - the step that had me lost for the longest time - understand to look for the problem at the `retainer` with the shortest distance. This is how I found that it was these `SIGTERM` event handlers that were being kept in memory. :)

fatso83:
Impressive! Glad to hear you worked it all out on your own. Those things
can be hard to track down.

Issue #76, opened 2015-09-12
Option for user agent


pocketjoso:
By default `Penthouse` sets a custom `Penthouse Critical Path CSS Generator` user agent. Some sites render differently based on UA; add an option to support using a custom UA.

pocketjoso:
Given that no one has ever asked for this functionality I'll close this. Can be added in the future if anyone needs it.

PR #109, opened 2016-03-16
add user agent option

nickdima:
Added the 'userAgent' option for specifying the user agent that phantom.js should use when loading the page. "Fixes" #76


pocketjoso:
Thanks! Will test, merge and deploy as soon as I get time - this weekend or sooner!

pocketjoso:
Great! :+1:

pocketjoso:
Published as `penthouse@0.9.2`. Thanks again!

Issue #113, opened 2016-03-21
Allow for configuring page render timeout


ndemengel:
I encountered the problem of pages not being completely rendered before generating the critical-path CSS. I've seen that there is a "TODO" comment regarding that point [in the code](https://github.com/pocketjoso/penthouse/blob/d4fec1d32ac9e95b98e6c8de49a65bc2c2974977/lib/phantomjs/core.js#L190).

It would be nice to have that timeout configurable. Or to wait for an element to be present in the page. More generally, we could also accept a predicate function that would tell when some expected condition is true before generating the CSS. This would be configurable as follows:

```js
penthouse({
    generateCssAfter: {  // once the page is loaded, wait for one of the following conditions before processing the css:
         delay: 1000,  // delay in milliseconds
         elementIsPresent: '#some .css selector',  // an element matching the CSS selector is present in the page
         condition: function () {  // or with a predicate (maybe not, that would involve (de-)serializing a function to pass it to phantom...)
             // ...
             return ready;
         }
    }
}, ...);
```

If you like the idea, I will submit a pull request in the next hours (update: I did submit pull requests to better illustrate).

pocketjoso:
Hi, sorry for my late answer. I've been thinking about this for a bit, as I'm not sure whether this is something `Penthouse` should handle.

> pages not being completely rendered before generating the critical-path CSS

Can you expand on this - when does this happen for you?

I remember adding the delay and the comment about it in the source code originally, but I think I did so for bad reasons. For example, content requiring JS to generate on the fly will not be included in the criticalcss - but also shouldn't be, as the first render happens before JS execution.

If you have valid use cases for handling this delay better I'm happy to merge your PR(s). Sorry for leaving them hanging for a while.

Cheers

ndemengel:
Well, thinking about it, I think you are right: it doesn't make a lot of sense, at least for most cases. The problem we have is for the following situations:
* A piece of JavaScript is included in the markup to be able to act very early and improve the placement of some elements that use a complex layout.
* Similarly, we have a banner that appears very early with an animation (for EU cookie law compliance), but we could delay that one a bit.

In those situations, the critical-path CSS doesn't contain the required CSS rules because its generation has been made too early ; but then when loading those pages from scratch, the JavaScript code will modify the DOM before the deferred loading of the non-critical-path CSS took place, so there is a time window where a visible part of the page is not properly styled. I suppose a solution would be to use forceInclude for the CSS classes involved, but it isn't always practical.

[...]

PR #115, opened 2016-03-21
Wait for condition before generation CSS

ndemengel:
Allow for configuring how penthouse wait for the page to be ready before generating the CSS, either by specifying a delay, or an element that should be present in the page:

```js
penthouse({
    generateCssAfter: {  // once the page is loaded, wait for one of the following conditions before processing the css:
         delay: 1000,  // delay in milliseconds
         elementIsPresent: '#some .css selector'  // an element matching the CSS selector is present in the page
    }
}, ...);
```


pocketjoso:
As we discussed in #113, this functionality should not be needed, rather should this timeout be removed completely from the source code.

PR #116, opened 2016-03-21
Wait for an arbitrary condition before generation CSS

ndemengel:
Going further than pull request #115, this one allows for waiting for an arbitrary condition to be true:

```js
penthouse({
    generateCssAfter: {  // once the page is loaded, wait for one of the following conditions before processing the css:
        condition: function () {  // an arbitrary predicate executed in the page context
            if (document.xxx()) return true
        }
    }
}, ...);
```


pocketjoso:
As we discussed in #113, this functionality should not be needed, rather should this timeout be removed completely from the source code.

PR #129, opened 2016-08-17
use os independent tmp directory

bezoerb:
Currently penthouse is not working on windows due to the hard-coded `/tmp` directory.
See: https://github.com/addyosmani/critical/issues/149


pocketjoso:
Oops - thanks! Will publish new version.

XhmikosR:
Do we really need yet another dependency? I don't see why https://nodejs.org/api/os.html#os_os_tmpdir wouldn't work.

PR #132, opened 2016-08-25
Add JSHint support

XhmikosR:
@pocketjoso: I went with the rules you already used, just removed any deprecated ones and tweaked the rest.

Let me know your thoughts and I'll rebase the branch to address any issues.

### DIFF
--- a/lib/phantomjs/core.js
+++ b/lib/phantomjs/core.js
@@ -4,11 +4,9 @@ var fs = require('fs')
 // phantom js related
 var system = require('system')
 var webpage = require('webpage')
-var jsesc = require('jsesc')
 
 var nonMatchingMediaQueryRemover = require('./non-matching-media-query-remover')
 
-var NORMALIZATION_DONE = 'NORMALIZATION_DONE'
 var GENERATION_DONE = 'GENERATION_DONE'
 
 var stdout = system.stdout // for using this as a file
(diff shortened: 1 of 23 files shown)


pocketjoso:
Hi thanks for the PR - well needed for this repo.

I actually never use jshint anymore, I use [standard](https://github.com/feross/eslint-config-standard) and eslint, but it's fine can update later.

I have some minor comments, otherwise it's good.

XhmikosR:
I don't use JSHint either. I just went with what you had.

Let's get this in and should make any changes later easier anyway.

lib/phantomjs/core.js
```diff
@@ -328,12 +280,12 @@ function getCriticalPathCss (options) {
   debuglog('stripped out non matching media queries')
 
   page.open(options.url, function (status) {
-    if (status !== 'success') {
```
pocketjoso: I prefer it like this ~ returning early; taking care of the edge case first before going to main code.

Actually looking at this now I would keep it like this but just return after the phantomExit below, then remove the unnecessary else clause and unindent that code.

package.json
```diff
@@ -15,14 +15,12 @@
...
-  "license": {
-    "type": "MIT",
-    "url": "github.com/pocketjoso/penthouse/blob/master/LICENSE-MIT"
-  },
+  "license": "MIT",
   "main": "./lib/index.js",
   "scripts": {
-    "test": "mocha --compilers js:babel-core/register test/core-tests.js",
-    "test-all": "mocha --compilers js:babel-core/register"
+    "jshint": "jshint .",
```
pocketjoso: can you name this command `lint` instead please?

pocketjoso:
Will merge after these two changes.

lib/phantomjs/core.js
XhmikosR: So you mean this?

``js
if (status !== 'success') {
  errorlog('Error opening url \'' + page.reason_url + '\': ' + page.reason) // jshint ignore: line
  phantomExit(1)
}

pocketjoso: I just checked the context and given how short the function is it doesn't matter to me. I'll merge as is. Otherwise what I meant was:
```
if (status !== 'success') {
  errorlog('Error opening url \'' + page.reason_url + '\': ' + page.reason) // jshint ignore: line
  phantomExit(1)
  return
}
debuglog('page opened')
page.evaluate(pruneNonCriticalCss, astRules, options.forceInclude, GENERATION_DONE)
```

PR #134, opened 2016-08-26
Remove unused devDependencies and normalize package.json

XhmikosR:
@pocketjoso: there seems to be more dev deps unused like express, morgan and portfinder. Should I remove those too?

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,9 @@
   },
   "license": "MIT",
   "main": "./lib/index.js",
+  "engines": {
+    "node": ">=0.12.0"
+  },
   "scripts": {
     "lint": "jshint .",
     "test": "npm run lint && mocha --compilers js:babel-core/register test/core-tests.js",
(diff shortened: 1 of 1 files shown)


XhmikosR:
Hmm, tests fails which is something I couldn't test since on Windows 10/12 tests are broken. I'll see if I can figure out the issue.

pocketjoso:
Still failing - let me have a look why.

XhmikosR:
Yeah, don't merge this yet, I'm still experimenting with it.

pocketjoso:
Ok, I'll wait then.

XhmikosR:
@pocketjoso: I can't figure it out. And since I can't test locally, it's harder for me to debug this.

So I reverted the deps update, just kept the other changes.

pocketjoso:
Okay, let's start with this and we can update deps later. 👍

PR #139, opened 2016-10-21
Optional unblock js requests

gavro:
Add blockJSRequests option (default = true).

Makes it possible to load JS files. See discussion here: https://github.com/pocketjoso/penthouse/issues/113
Made it optional. Choice = good, even if it might be wasteful :)


pocketjoso:
Sure, code looks good.

Just update the README to list the option you added, and then I'll merge. 👍

gavro:
Awesome, thnx. Added to README. Do you need a squashed commit or are these commits fine?

pocketjoso:
We could squash the lint error commit into the one it fixed - but not super important, will merge now as is.

Thanks for the PR!

pocketjoso:
Published as `0.9.16`.

Issue #140, opened 2016-10-25
Update "css" URL


qasimzee:
Seems like "git://github.com/pocketjoso/css.git" has bin invalid. Can we please update it from "git://github.com/pocketjoso/css.git" to "https://github.com/pocketjoso/css.git"

pocketjoso:
Hi, where do you see this link listed?

qasimzee:
Thanks for very fast response.

Line 30, https://github.com/pocketjoso/penthouse/blob/master/package.json

pocketjoso:
Just quick tip for future - you can highlight lines via url on github (also via clicking on them):
https://github.com/pocketjoso/penthouse/blob/master/package.json#L30

pocketjoso:
Can you tell me more before I make this change - what problem does this invalid bin cause for you? Does `Penthouse` not run?

qasimzee:
During installation on windows, it throws an error.

qasimzee:
Following is the error:

npm WARN addRemoteGit Error: Command failed: git -c core.longpaths=true config --get remote.origin.url
npm WARN addRemoteGit
npm WARN addRemoteGit     at ChildProcess.exithandler (child_process.js:213:12)
npm WARN addRemoteGit     at emitTwo (events.js:87:13)
npm WARN addRemoteGit     at ChildProcess.emit (events.js:172:7)
npm WARN addRemoteGit     at maybeClose (internal/child_process.js:827:16)
npm WARN addRemoteGit     at Socket.<anonymous> (internal/child_process.js:319:11)
npm WARN addRemoteGit     at emitOne (events.js:77:13)
npm WARN addRemoteGit     at Socket.emit (events.js:169:7)
npm WARN addRemoteGit     at Pipe._onclose (net.js:477:12)
npm WARN addRemoteGit  git://github.com/pocketjoso/css.git resetting remote C:\Users\qzeeshan\AppData\Roaming\npm-cache\_git-remotes\git-github-com-pocketjoso-css-git-45193baf because of error: { [Error: Command failed: git -c core.longpaths=true config --get remote.origin.url
npm WARN addRemoteGit ]
npm WARN addRemoteGit   killed: false,
npm WARN addRemoteGit   code: 1,
npm WARN addRemoteGit   signal: null,
npm WARN addRemoteGit   cmd: 'git -c core.longpaths=true config --get remote.origin.url' }
[...]

PR #141, opened 2016-10-26
Update package.json

qasimzee:
Updating css package url. Cause issue #140 in windows 10


pocketjoso:
Thanks for PR, will just test that it works tonight, and them merge and ship!

PR #161, opened 2017-05-13
Code refresh

pocketjoso:
### Feature-wise:
* Added `Promise` support 🎉 (about time 😝)

### Cleanup:
* No more sync IO operations inside penthouse (stop blocking node process)

### Development
Finally transpiling code via babel (hence new `src` folder), so can use modern JS (async/await) w/o sacrificing support for older node.

Also added:
- [yarn](https://github.com/yarnpkg/yarn) (deterministic builds),
- [standard](https://github.com/feross/standard) (replacing jshint for better, config-free, JS linting),
- [prettier](https://github.com/prettier/prettier) (auto code-formatting)
Open source FTW!  🌈


pocketjoso:
Spent a lot of time today looking at why it seemed like these changes were causing a perf regression - was seeing on average about 0.5 - 1s slower job execution per `penthouse` invocation - however in the end seemed to only happen when overloading system with too many parallel calls (~more than number of CPU cores)... so as there seems to be no (significant) negative effect on a small and more normal amount of parallel calls,  will merge and continue to monitor for perf changes.

PR #163, opened 2017-05-16
Make babel-polyfill Regular Dependency

corneliusio:
Had some issues sometime after the 0.11.0 update because babel-polyfill couldn't be found. Should just need to be moved from dev deps to regular deps.

Let me know if I'm mistaken or missed anything.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
   },
   "dependencies": {
     "apartment": "^1.1.1",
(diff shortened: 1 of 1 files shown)


pocketjoso:
You're right, thanks for the PR!

pocketjoso:
published as `penthouse@0.11.3`! 🎉

PR #166, opened 2017-06-09
Replace babel-polyfill with regenerator-runtime

corneliusio:
So, for kicks and giggles I was trying out Node 8.0 on my machine (I know, I know… it's not even LTS yet), but found that everything was working fine with my build system except for penthouse. So dug in and found that the only issue was babel-polyfill seemingly not working, and that since really the only reason it was (apparently) being pulled in for was `regeneratorRuntime`, using the regenerator-runtime was an easy fix.

It also has the added bonus of being much lighter weight, since you don't have to pull in all of the extra polyfill stuff. And as an added bonus, you look to be set for Node 8.0 with this change.

Also, was having some issues with my own .eslintrc rules bubbling up into the prettier-standard call and adding semicolons to everything—so also added a local .eslintrc file which will be helpful for any other contributors. Wasn't sure what you're rules look like, so simply added the only one that was needed to keep my rules from changing any of your styles.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,9 @@
+{
+    "env": {
+        "node": true,
+        "es6": true
+    },
+    "rules": {
+        "semi": "never"
+    }
+}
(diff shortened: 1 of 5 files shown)


pocketjoso:
Thanks for the PR! seems good, did you try it on Node 4?

corneliusio:
Yeah, that was the one thing (ironically) I didn't make absolutely sure about. The rest of my build system doesn't run on 4 so it wasn't trivial to simply switch and run it. I can def make sure though, will get back to you when I know for sure.

corneliusio:
Happy to say that it runs in Node v4.8.3 perfectly fine!

pocketjoso:
👍

pocketjoso:
Released in `v0.11.4`.

PR #167, opened 2017-06-09
Assign Local regeneratorRuntime Variable

corneliusio:
So, interestingly enough, everything worked with the local copy before but after pushing everything to npm and using the published version, it seems that for some reason that regeneratorRuntime is not showing up in the global scope as expected. 

Originally didn't assign it locally because prettier-standard was throwing a no-unused-vars error (prob just because of my own eslint rules), but just disabled that error for that line and actually tested it by pulling from my repo after instead of linking to the local copy.

Sorry for the faulty PR previously. 😓


pocketjoso:
Hmm I can't quite see why that should be necessary, see f.e. the (very sparse) documentation here:
https://github.com/facebook/regenerator/tree/master/packages/regenerator-runtime.

Are you sure the problem is not something else?

corneliusio:
Yeah, honestly it doesn't make any sense to me. Running into to some very odd inconsistencies over here. I'll keep digging into it, it's certainly possibly that there's some reason it's acting wonky on my computers related to changing the node versions. I'm actually having the same `ReferenceError: regeneratorRuntime is not defined` on other releases of the package now while also using either node 6 or 8 at the moment. 

As long as it's working on your end, I say you can ditch this PR. There really shouldn't be any reason we have to define the regeneratorRuntime var locally (which is why I didn't originally) so I'm going to trump it up to user error on my end at the moment.

pocketjoso:
Yeah, it is working for me in node 8, so closing.

PR #174, opened 2017-08-20
1.0.0: Chrome headless

pocketjoso:
This is a bigger change, so once it's done I plan to release it as `1.0.0`, partially also to reflect that Penthouse is a mature library that can be used in production.

**You can test this PR using `penthouse@next`.**
Feedback welcome! (please make sure you're on the latest `next` tagged version, and report the exact version, os and node version - thanks!)

----

Migrating from Phantomjs (which is [no longer maintained](https://groups.google.com/forum/#!topic/phantomjs/9aI5d-LDuNE) to running on headless chrome, via the official(?) [puppeteer](https://github.com/GoogleChrome/puppeteer) api.

## Breaking changes
* drops support for `node < 6.4`

## Improvements
* **In general Penthouse is now about twice as fast! 🎉 🏇**
* Test execution is ~ **4x faster** 🏎 
* parallel execution is run in the same browser instance (in different tabs), instead of in different browser instances, consuming less resources
* before/after screenshots support 📷, making it easy to validate that the critical css is correct.

## Other changes
* support for `cssString` - if you already have the css in memory, no need to pass a file path (old `css` attribute)

I plan to make these further improvements before releasing:
- [x] speed up execution when the original css cannot be parsed (i.e. the normalization process), by running it in the same browser instance as the general generation)
- [x] speed up execution further by re-using the same chromium browser instance when possible, rather than starting multiple separate ones (for separate Penthouse calls)
- [x] add an option to generate before/after screenshots

**TODO musts before merge**
- [x] fix for Chromium instances not always (?) being terminated, ending up hogging resources
- [x] fix for new timeouts, not happening with PhantomJS
- [x] make use of again or drop: `userAgent` option
- [x] make use of again or drop: `customPageHeaders` option
- [x] kill any running browser instance on sigkill etc
- [x] fix cases of css files causing page.eval calls to hang
- [x] fix new memory leaks (if remains after above)
- [x] extra errors thrown following timeouts or page crashes -
 will be fixed via https://github.com/GoogleChrome/puppeteer/issues/695

** Non blockers - known hard-to-reproduce problems
- [x]  `UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): Error: Protocol error (Target.closeTarget): Target closed.` - When running long sequence of jobs? (via critical)
- [ ] some pages never load in puppeteer (with JS disabled) - issue here https://github.com/GoogleChrome/puppeteer/issues/662
- [ ] callback never called if chrome or is it page(?) crashes (due to being out of memory) https://github.com/pocketjoso/penthouse/pull/174#issuecomment-326922920
- [ ] unwanted multiple chrome instances are created (and kept alive) in some circumstances (probably related to `unstableKeepBrowserAlive`)


pocketjoso:
Some issue with used Chromimum instances now being terminated, ending up hogging resources - needs to be resolved before release.

wardpeet:
@pocketjoso does it happen on all linux machines? Or only specific distros? Which version of chromium?

pocketjoso:
@wardpeet By now it's been working for me on Linux ubuntu, and macOS. I haven't tested any other OS or distros.

pocketjoso:
@wardpeet By now it's been working for me on Linux ubuntu, and macOS. I haven't tested any other OS or distros. I updated the readme with what I needed to do to get it working:
https://github.com/pocketjoso/penthouse/blob/91e117dfd2a6286db3c007752cb93151c9888066/README.md#not-working-on-linux

pocketjoso:
Okay, this looks in good shape to be merged now. Will look to do so next week.

soukicz:
Hi, I have been testing this version and results are better than with PhantomJs. But sometimes I get following error. I wasn't yet able to localize it in more detail.
image

pocketjoso:
@soukicz I've seen this error too, I have some handling for it inside Penthouse to try to just restart the job. I don't know yet what causes it, my first guess would be the process ran our of memory..

Can you share more details on how you called `Penthouse`? Multiple calls at the same time?

soukicz:
It was memory - I was running multiple instances in parallel. Crashing on low memory isn't problem (at least not a problem of penthouse). What is problematic is error handling - I am using callback version (not promises) and callback is never called on chrome crash.

pocketjoso:
Yep that's a bug - I'll make sure to fix it before releasing this.

reznord:
This is awesome. I was tracking this from https://github.com/addyosmani/critical/issues/229.

soukicz:
I have already deployed this version to production and (except that crash handling) it's working perfectly!

pocketjoso:
@soukicz Good to hear. Same on my end, quite stable now aside from that error. Trying to narrow it down - can you tell me some rough size of the (original) css you passed in to Penthouse? Was it large (several MB)? At roughly how many parallel calls did you get the crashes?

soukicz:
It was this page and its CSS - https://www.servisnidily.cz/ (https://e510.ecdn.cz/t/8s/6w5i5srxqskk/css/package.min.css) - it was running in paralel on 6 subpages on AWS EC2. I am not sure how much memory was available but probably ~2GB. 
Chrome crash is fine (it is huge and it runs in multiple instances). It just should handle the promise rejection.
I was dealing with similar problem with puppeteer and I will try to find it in penthouse source code, but JS isn't my strongest language :)

pocketjoso:
@soukicz Thanks, this helps. I will get to it, sometime this week.

reznord:
@pocketjoso is it possible to use only one instance of chrome and multiple tabs (instead of multiple chrome instances). That might speed up the tests in penthouse.

I was actually looking for a way to implement it in [preact-cli](https://github.com/developit/preact-cli).

soukicz:
@reznord openning new tabs is indeed a lot faster then starting new chrome instance. But it might be even more useful to have config attribute to pass url of shared Chrome instance outside of penthouse.

We are already using Puppeteer to generate PDFs and it would safe a lot of memory to just reuse this already running instance. I was going to play with this after release of penthouse 1.0.0

pocketjoso:
@reznord @soukicz Penthouse will never open separate browser instance anymore, since this PR. It does open new jobs in new browser tabs, and by default only closes the browser when there are no more penthouse calls running. Hence, you can keep the same browser instance running by doing calls in parallel.

If this is not good enough for you, you can use the `unstableKeepBrowserAlive` setting, which always re-uses the same browser. Note though that this API is not stable, and might change in the future. It also _never_ closes the browser - the only way to close it is by shutting down the node process. Just with caution ⚠️

reznord:
@pocketjoso Can you point me out the exact code for this from the commit?

pocketjoso:
@reznord https://github.com/pocketjoso/penthouse/pull/174/commits/1ded7ac4c70aa139fa326dbf486b76886f0a3f72

pocketjoso:
@soukicz The latest penthouse version now (`penthouse@1.0.0-rc13`) took some updates from puppeteer that should fix these uncaught errors. Can you check again now please?

soukicz:
Didn't run in that issue again.

soukicz:
I got some errors again, but it is burried in app and I am not able to reproduce them again. It is on R13)

```
UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): AssertionError: INTERNAL ERROR: failed to find request for interception redirect.
DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
It happened on this url: https://repo-488-8a-tpltest.simplia.cz/

pocketjoso:
Thanks for reporting @soukicz, will investigate.

soukicz:
I have noticed debuglog calls in source - how can I enable this output to get more detailed logs about crashes?

soukicz:
It seems that my errors were caused by recent updates to puppeteer. It is working again after downgrading to https://github.com/GoogleChrome/puppeteer/commit/f11e09d0d413a42c177cac4b1b0a297d98b5fa17

pocketjoso:
> I have noticed debuglog calls in source - how can I enable this output to get more detailed logs about crashes?

set `penthouse.DEBUG = true` after importing it. It might not be too helpful for you in this case as it doesn't currently give you any puppeteer log info,. It will just show you how far into penthouse own execution you got.

pocketjoso:
@soukicz regarding the ` AssertionError: INTERNAL ERROR` for https://repo-488-8a-tpltest.simplia.cz/ - it was only a sporadic error, or? I looked through my own logs (20,000 or so calls to Penthouse) and no such errors on the latest version, not when I try on this url either. If it's really rare could just be that it didn't happen to me yet.

soukicz:
The assertion errors happen sometimes on redirects. I was giving it redirected URL but after I fixed this, it was still occurring (~1 in 5 runs) - probably because of some resource redirect?

Now it crashes in 50% of runs – sometimes it is that assertion error and sometimes just timeout. I am now trying older and older puppeteer versions (commit by commit). It was all working perfectly until docker image was rebuilt with new puppeteer version yesterday. Nothing changed after reverting penthouse to R9 and some recent change in puppeteer is currently my best suspect.

image

pocketjoso:
@soukicz Ideal then would be to create an issue on puppeteer, if it can be isolated. I have also seen some changes (mostly for the better, but some possible regressions) since https://github.com/GoogleChrome/puppeteer/commit/0bea42bd8c4002ed28d100d9ccc06c6ac03a9b70 and https://github.com/GoogleChrome/puppeteer/commit/e5c17eecb93b995c78515e8b1b0df13c392efd8e - but it sounds like you're going even further back in the history? Let me know what you find.

soukicz:
Would it be possible to add some way to set executablePath? As described in
https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker

I have tried to create PR but I didn't managed to make it work...

pocketjoso:
Should be possible yes. Can you open the PR and I can see how far you got,
and help you land it?

reznord:
Can we test this out using the rc version?

pocketjoso:
Just published what should be the final version before merging to `master`:
`penthouse@1.0.0-rc14`

This fixes some issues with leaking `Target closed` errors from puppeteer, that were due to a missing `await` on my end for the `page.close()` call. Penthouse now also handles recovering if the browser instance gets shut down/crashes, which is helpful for the `unstableKeepBrowserAlive` mode.

cc @reznord

reznord:
Amazing 😍, will test it out on critical now and see if any errors pop out.

pocketjoso:
Let's do this! 🚀

bezoerb:
\o/

reznord:
Wohooo \o/

pocketjoso:
Posted on medium about the release:
https://medium.com/@pocketjoso/penthouse-1-0-official-release-ece995b0d29e

PR #176, opened 2017-09-05
Check for node version in postinstall script

### LINKED ISSUE #175: Check for node version on postinstall script
Right now we are not checking for node version. It would be better to have a check where we throw an error if the node version is less than the required (specified in `engines` in package.json)

reznord:
This closes #175 (added chalk as a dependency)

### DIFF
--- a/check.js
+++ b/check.js
@@ -1,0 +1,22 @@
+const colors = require('chalk');
+const pkg = require('./package.json');
+
+const version = parseFloat( process.version.substr(1) );
+const minimum = parseFloat( pkg.engines.node.match(/\d+/g).join('.') );
+
+module.exports = function () {
+  if (version >= minimum) {
+    return true;
+  }
+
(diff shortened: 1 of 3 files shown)


soukicz:
Is this really needed? Penthouse has puppeteer as dependency and it already checks node version. It shouldn't be possible to run `npm install` even without post-install script.

reznord:
That's true. But we may not be using puppeteer in the future. So, it is always best to add it in the code (that's what I think).

pocketjoso:
After sitting on this for a while I must say I agree with @soukicz - Puppeteer is really the bottleneck for the minimum version of Penthouse, and does handle checks. It just adds extra work keeping this up to date, as now when Puppeteer [changed their supported Node versions](https://github.com/pocketjoso/penthouse/pull/174/commits/810d788b87d9931efebcad05f99c8a14446a4e58) again. So thanks for the work, perhaps in the future!

PR #177, opened 2017-09-14
set puppeteer executable path

soukicz:
It seems to be working but I am not sure if this is the right way to do it.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -273,7 +273,7 @@ const m = (module.exports = function (options, callback) {
   return new Promise(async (resolve, reject) => {
     // still supporting legacy callback way of calling Penthouse
     const cleanupAndExit = ({ returnValue, error = null }) => {
-      if (browser) {
+      if (browser && !options.unstableKeepBrowserAlive) {
         if (_browserPagesOpen > 0) {
           debuglog('keeping browser open as _browserPagesOpen: ' + _browserPagesOpen)
         } else {
--- a/src/index.js
+++ b/src/index.js
@@ -20,7 +20,7 @@ const DEFAULT_BLOCK_JS_REQUESTS = true
 let browser = null
 let _browserLaunchPromise = null
 // browser.pages is not implemented, so need to count myself to not close browser
-// until all pages are closed again
+// until all pages used by penthouse are closed (i.e. individual calls are done)
 let _browserPagesOpen = 0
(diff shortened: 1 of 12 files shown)


pocketjoso:
Hmm I didn't give you clear directs, I was thinking more of making it a regular Penthouse option, rather than a hidden `m.<debug_setting>` flag. But we could start here and I can move it later if there are more requests for it. I will comment on your code too now, since it won't work the way you wrote it.

src/index.js
```diff
@@ -267,6 +271,7 @@ const generateCriticalCssWrapped = async function generateCriticalCssWrapped (
 const m = (module.exports = function (options, callback) {
   // init logging and debug output
   normalizeCss.DEBUG = m.DEBUG
+  const PUPPETEER_EXECUTABLE_PATH = m.PUPPETEER_EXECUTABLE_PATH
```
pocketjoso: if you declare a const here like this, it will not be accessible in the `launchBrowserIfNeeded` function.
Can you for now just declare this variable as a `let` on line 29 (without any assignment, i.e. `let PUPPETEER_EXECUTABLE_PATH`), to make it accessible in the scope of `launchBrowserIfNeeded`, then on this line just assign the value (`PUPPETEER_EXECUTABLE_PATH = ....`).

Then should work for now, and I will refactor this code a bit in the future.

pocketjoso:
You would now need to rebase this PR too, and make sure you drop all previous commit that I changed upstream, and only re-apply the one that's yours.

Saturate:
We need this possibility as well, I'm thinking an Environment Variable. Might make another PR with the possibility of this.

pocketjoso:
@Saturate a new PR for this could be a great option.

Saturate:
@pocketjoso I'm now thinking it might be better in the Puppeteer repo, as that would make sense I guess.

pocketjoso:
@Saturate Yeah you're right, then no change would be needed in Penthouse.

Saturate:
Yep @pocketjoso I started this https://github.com/GoogleChrome/puppeteer/issues/1335 and will make a PR if they agree. If they don't I'll make a PR for you :-)

pocketjoso:
Closing because of inactivity, but let's resume the discussion (and re-activate the work) if the problem is still here.

PR #179, opened 2017-09-19
Changes package.json to point to Puppeteer's package instead of repo. Also fixes typo on README.md

mAiNiNfEcTiOn:
# Problems
* `README.md` has a small typo in the example (an extra `},`, not needed 😉 ).
* The fact that the project uses `puppeteer` is awesome. However, because it is pointing to Google's github repo it gets the master and does not run the build to be compliant with Node 6, which causes the following error:

screen shot 2017-09-19 at 17 17 27

# Solutions/Work Done:

* Fixes the typo on the `README.md`'s example.
* Changes the `package.json` to point to the (currently latest) puppeteer's package version instead of its repo. Also changes the `yarn.lock` to point to the correct version. Both steps done via `yarn add puppeteer` as @pocketjoso suggested. 

Note: this is my first PR in this repo, all feedback about contribution procedures (but not only) is more than welcome 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,6 @@ penthouse({
     userAgent: 'Penthouse Critical Path CSS Generator', // specify which user agent string when loading the page
     renderWaitTime: 100,            // ms; render wait timeout before CSS processing starts (default: 100)
     blockJSRequests: true          // set to false to load (external) JS (default: true)
-    },
     customPageHeaders: {
       'Accept-Encoding': 'identity' // add if getting compression errors like 'Data corrupted'
     },
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "css-fork-pocketjoso": "^2.2.1",
     "css-mediaquery": "^0.1.2",
     "jsesc": "^1.0.0",
-    "puppeteer": "GoogleChrome/puppeteer"
+    "puppeteer": "^0.11.0"
   },
   "devDependencies": {
     "babel-cli": "^6.24.1",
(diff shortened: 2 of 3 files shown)


pocketjoso:
Hi there, thanks for the PR. In general I absolutely agree with you, but there are some critical commits that have been pushed in puppeteer since `0.10.2` that Penthouse needs (otherwise we have errors thrown by Puppeteer that cannot be caught and leads to warnings in Node since they will in future versions cause it to crash).

So sorry, we need to wait until `0.11.0` lands. Happy to merge by then.

mAiNiNfEcTiOn:
@pocketjoso thanks for the feedback. Totally understand, but then that forces me, on my package.json to add puppeteer as a dependency and do `rm -Rf node_modules/penthouse/node_modules/puppeteer` to be able to have the package ready for node 6. But I guess that's ok until we have puppeteer's v0.11.

Do you want me to extract the README's fix to another PR?

mAiNiNfEcTiOn:
@pocketjoso Updated the PR to point to v0.11.0 as a minimum for `puppeteer`.

pocketjoso:
Not sure how I missed that you did update the `yarn.lock` file - was just going via the commit message. Will merge - thanks again for the PR!

pocketjoso:
Available as `penthouse@1.0.2`

mAiNiNfEcTiOn:
@pocketjoso you didn't miss it. I just made the changes as you requested and rebased 😉 thanks for the merge

PR #182, opened 2017-09-21
Test node 6

bezoerb:
Should test the min node version required

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: node_js
 node_js:
-    - "7.10"
+    - "6.4"
+    - "7"
+    - "8"
 dist: trusty
 addons:
   apt:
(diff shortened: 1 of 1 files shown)


pocketjoso:
Indeed, thanks for PR.

I re-ran the tests now after taking up Puppeteer@0.11 in #179, and got a different error, but still same cause.

It works locally for me on Node 6, it seems to only be a problem in Travis.. 
[Puppeteer specifically run the build command node 6 in their travis file](https://github.com/GoogleChrome/puppeteer/blob/master/.travis.yml#L19-L26), should every library that uses it be required to add the same configuration?

I don't know the right answer here. @bezoerb do you want to try running the puppeteer build command from the travis file?

bezoerb:
Tested critical with node6 locally (node_modules removed) and everything worked fine. So I guess the travis tweak is the right way to go here :)

pocketjoso:
Looks good, merging!

PR #186, opened 2017-09-24
fixed options.blockJSRequests always getting default value (true)

alonp99:
Bug: trying to set `blockJSRequests` as false fails, unwanted behavior: fallback to the default value (true).


pocketjoso:
You're right, thanks for the PR!

pocketjoso:
Pushed as `1.0.4`.

PR #199, opened 2017-11-10
Support passing puppeteer browser in options

### LINKED ISSUE #198: Passing puppeteer browser in options
Sometimes to generate critical css I need to do a couple of things (e.g. login user).

I'm not sure what's the best way to do that, but it would be great if I can create puppeteer browser, do some stuff then pass it to penthouse.

If that makes sense then I can make a PR for that?

bitriddler:
Checking if `puppeteer.getBrowser` function is passed in props to use it for launching browser instead, closes #198.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -35,9 +35,12 @@ let _browserLaunchPromise = null
 // browser.pages is not implemented, so need to count myself to not close browser
 // until all pages used by penthouse are closed (i.e. individual calls are done)
 let _browserPagesOpen = 0
-const launchBrowserIfNeeded = async function (debuglog) {
+const launchBrowserIfNeeded = async function (options, debuglog) {
   if (browser) {
     return
+  }
+  if (options.puppeteer && typeof options.puppeteer.getBrowser === 'function') {
+    _browserLaunchPromise = Promise.resolve(options.puppeteer.getBrowser())
   }
   if (!_browserLaunchPromise) {
     debuglog('no browser instance, launching new browser..')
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -265,7 +269,7 @@ const generateCriticalCssWrapped = async function generateCriticalCssWrapped (
         } else {
           console.log('restarting chrome after crash')
           browser = null
-          await launchBrowserIfNeeded(debuglog)
+          await launchBrowserIfNeeded(options, debuglog)
```
pocketjoso: it's not safe to try to re-use the browser you launched in this case, as it (chrome) has crashed.

pocketjoso: The easiest way forward with the logic you set up would be to pass `{}` instead of options here. I might clean up the internal logic later, but should be good enough for now.

pocketjoso: However that would mean that the browser that restarts does not have the setup you require.

pocketjoso:
Thanks for the PR.
It's a bit problematic to just pass in a browser instance, as penthouse automatically tries to restart Chrome should it crash, and this process is not transparent to your code..

To not complicate things, I think the easiest solution is to disable the auto-restarting when the user passes a `puppeteer.getBrowser` function. The auto-restarting is mostly useful when you try to batch generate a really large number of critical css files, which is not the most common use case, so should be ok to ignore until further.

I'll place comments in the code to follow this approach up.

src/index.js
```diff
@@ -265,7 +269,7 @@ const generateCriticalCssWrapped = async function generateCriticalCssWrapped (
         } else {
```
pocketjoso: I would now change this to:

src/index.js
```diff
@@ -369,7 +373,7 @@ const m = (module.exports = function (options, callback) {
       return
     }
 
-    await launchBrowserIfNeeded(debuglog)
+    await launchBrowserIfNeeded(options, debuglog)
```
pocketjoso: could you change this to:

src/index.js
```diff
@@ -35,10 +35,13 @@ let _browserLaunchPromise = null
 // browser.pages is not implemented, so need to count myself to not close browser
 // until all pages used by penthouse are closed (i.e. individual calls are done)
 let _browserPagesOpen = 0
-const launchBrowserIfNeeded = async function (debuglog) {
+const launchBrowserIfNeeded = async function (options, debuglog) {
```
pocketjoso: then in line with the other two changes I asked for, this would be updated to:

pocketjoso:
With these changes - as long as the tests pass - this should be good to go for me.

pocketjoso:
Finally, can you update the README to mention this feature?

bitriddler:
@pocketjoso Thanks for your review, I updated the PR, please take a look 👍 

Also I've tested the new build in my project (with and without getBrowser option)

package.json
```js
...
    "penthouse": "git+https://github.com/bitriddler/penthouse.git#build-getBrowser-option",
...
```

and everything seems ok

PR #203, opened 2017-11-27
Update puppeteer

Saturate:
puppeteer 0.13 has support for the `PUPPETEER_DOWNLOAD_HOST`
configuration, this is useful for people behind proxy.

Would be nice with a new `1.1.3` version of penthouse.


pocketjoso:
Looks good, thanks for the PR! My only nitpick is that the `lib` files should not be checked into git; must be an issue with the ignore entry in the `.gitignore`. Can you at least remove the `lib` file from your commit, or even better also fix the `.gitignore`?

Then I'll merge and I'll push a new penthouse version.

Saturate:
Ops. The `.gitignore` looks fine, must have been my VS Code or something that did something strange when I renamed the API call. It should be gone again.

You might wan't to update the `package-lock.json` I can't do that here, due to the proxy. But that will need to happen if you bump the version of penthouse anyway I guess.

pocketjoso:
Thanks for the update.

I don't personally use `package-lock.json` as I use `yarn`.

pocketjoso:
Published as `1.1.3`.

Saturate:
That's awesome, thanks for the quick support :-)

Issue #204, opened 2017-11-29
Extracted CSS in v1.1.3 is very different from that extracted in v1.1.2 usign local file in url parameter(file://)


topogigiovanni:
I'm having very different results after update to v1.1.3. The generated file is considerably different and smaller.

I also noticed is the screenshot can not parse HTML when the url parameter contains a file (file: //)

Below, we have the result of the screenshot of the same code using different versions of the Penthouse.

My Node version is v6.10.3

in v1.1.2
![screenshot-v1-1-2](https://user-images.githubusercontent.com/5891618/33367738-b2140c34-d4d6-11e7-99ff-fd2f7b114a5e.jpg)

in v1.1.3
![screenshot-v1-1-3](https://user-images.githubusercontent.com/5891618/33367750-beae7b00-d4d6-11e7-91f1-a52e9251d4cd.jpg)

pocketjoso:
Hi Giovanni,
Where did you get the second screenshot from? that looks like html, which if that's what you're seeing in your browser should have nothing to do with whether you are using critical css or not.

Do you have a public url for your site that I can test with myself?

Saturate:
For me it seems like puppeteer opens the correct file, but views it a html. Pretty strange, I did not happen when I tested the bump of puppeteer to 0.13, but happens on a fresh install.

Saturate:
Here is a screen capture, the one opens by the API is just HTML. The other one is one I opened. It's opened in `headless: false` mode.

![bug](https://user-images.githubusercontent.com/453018/33377251-17cba0ac-d511-11e7-82fe-eb68e577432d.gif)

topogigiovanni:
Hey guys,
What's happening here is the same thing that @Saturate has shown. But I'm not testing directly on the Puppeteer. But apparently the Puppeteer is not being able to interpret the HTML correctly even coming from a file, causing the error of the extraction of the critical CSS.

I published the tests I'm doing here https://github.com/topogigiovanni/penthouse-diff
remembering that I'm using version 6.10.3 of the Node

Saturate:
I tried running it on puppeteer directly with the following code, and it works:

```
const puppeteer = require('puppeteer');

(async () => {
    const browser = await puppeteer.launch({
        waitUntil: 'networkidle2',
        ignoreHTTPSErrors: true,
        args: ['--disable-setuid-sandbox', '--no-sandbox'],
        headless: false
    });
    const page = await browser.newPage();
[...]

PR #206, opened 2017-11-30
 Use debugjs for debugging

Saturate:
Uses debugjs for debugging, the same as puppeteer is using, this allows
for easier debugging, in a way that's similar with puppeteer. Also document this change.

Here is a preview from powershell:

image

Note:

Example could also be `env DEBUG="penthouse*" node script.js` but I feel the other is easier to understand. We could also scope everything in `index.js` to `penthouse:index` then `env DEBUG="penthouse:*" node script.js` would work fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,15 @@ The Penthouse Node module can also be used in Gulp.
 
 ## Troubleshooting
 
+
+### Enable debug logging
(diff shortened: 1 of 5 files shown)


pocketjoso:
Looks good to me - thanks for pr! Just fix the linting that is causing the test to fail, then I'll look to merge this.

Saturate:
Ooops, I missed that one. It's now fixed.

pocketjoso:
Awesome! Will most likely hold of with new releases until we have a fix for #204.

Saturate:
Makes sense, it's also just debugging nothing that really warrens a new release :-)

PR #209, opened 2017-12-03
New fault-tolerant ast parser (css-tree)

### LINKED ISSUE #208: fault tolerant CSS AST parsing
This has long been a problem for penthouse, as the current [css](https://github.com/reworkcss/css) parser is not fault tolerant, and (in best case) throws when there's anything in the input css it doesn't understand.

The goal here would be to switch to an ast parser that is:
- as fast as possible
- as fault tolerant as browsers (i.e. [the spec](https://www.w3.org/TR/css-syntax-3/#error-handling))

Promising alternative parsers are:
- [csstree](https://github.com/csstree/csstree) @pocketjoso looking into it, will open PR soon
- [CSSOM](https://github.com/NV/CSSOM)
- [postcss](https://github.com/postcss/postcss) + [postcss-safe-parser](https://github.com/postcss/postcss-safe-parser) @bradlc has something working, will open PR

(this list was found via [ast explorer](https://github.com/fkling/astexplorer/blob/master/README.md))

pocketjoso:
Should resolve #208. As discussed in there however there are other css parsers out there and we might see more PR's coming in for other ones - if so I'm willing to switch to whichever one yields the best results - most importantly accuracy in critical css generation and speed of execution.

This PR is available to test as `penthouse@1.2.0-rc` (or `penthouse@rc`).

Regarding the css-tree usage, as mentioned in one commit:
NOTE: using non recommended way of converting css-tree's Object based
List representation of AST into standard JS arrays because easier to work with,
but as discussed might change this in the future:
csstree/csstree#69

Needs some more solid testing before considered to merge.

*Needs follow up action:*
* `propertiesToRemove` is via this PR temporarily put inside a try/catch, and will not be (successfully) applied to faulty (otherwise now fully handled) css. [More info](https://github.com/pocketjoso/penthouse/pull/209/files#r154793762)


pocketjoso:
Have done some serious testing today and this version looks very solid, just a straight up improvement over previous ast parser. Will give it one more day, looking to merge in ~24h.

pocketjoso:
What are you confused about @bezoerb? Let me know if you have any questions I can answer.

src/postformatting/index.js
```diff
@@ -1,43 +1,50 @@
...
+  debuglog('propertiesToRemove')
   // remove irrelevant css properties
-  finalCss = apartment(finalCss, {
-    properties: propertiesToRemove,
-    // TODO: move into pruneNonCriticalCss script
-    selectors: ['::(-moz-)?selection']
-  })
-  debuglog('postformatting: cleaned css via apartment')
+  try {
+    finalCss = apartment(finalCss, {
```
pocketjoso: This call to apartment is the only part of Penthouse that still uses the non-fault tolerant rework css parser. Now that the rest of the pipeline is fault tolerant, this call here can be made with css with errors inside, causing rework css’ ast parsing to throw.

Hence my quick fix here is to try/catch and essentially classify this property removal as an "enhancement" for now. This will have the side effect of the critical css increasing in size - only for css files with errors inside. In my tests across thousands of critical css generations This lead to on average about a 0.5% size increase, so nothing dramatic.

Still, this functionality will return. I pinged the author of apartment to give him a chance to say whether he is interested in updating his library, otherwise I'll just rewrite that code and bring it in to penthouse natively - but this would be in another week or so.

bezoerb:
No confusion @pocketjoso, guess i just hit the wrong emoji on my phone ;)
Awesome work 👍

src/browser-sandbox/pruneNonCriticalCss.js
```diff
@@ -132,40 +127,53 @@ export default function pruneNonCriticalCss ({
   }
 
   function isCssRuleCritical (rule) {
-    if (rule.type === 'rule') {
+    if (rule.type === 'Rule') {
       // check what, if any selectors are found above fold
-      rule.selectors = rule.selectors.filter(isSelectorCritical)
-      return rule.selectors.length > 0
+      const selectors = rule.prelude.value.split(',')
```
lahmatiy: That's incorrect approach, since pseudo classes, attribute selector value and comments can contain a comma (e.g. `:not(a, b) /* comme,nt */ [foo="hello, world"]`).
It's better to parse a selector and filter on tree traversal.

pocketjoso:
On 5 Dec 2017 02:11, "Roman Dvornov" <notifications@github.com> wrote:

*@lahmatiy* commented on this pull request.
------------------------------

In test/util/normaliseCssAst.js
<https://github.com/pocketjoso/penthouse/pull/209#discussion_r154819384>:

>  // because dont want to fail tests on white space differences
-export default function normaliseCssAst (cssString) {
-  return css.parse(css.stringify(css.parse(cssString), { compress: true }))
+export default function normaliseCssAst (css) {
+  return parseCssAst(

Did you try to use a single parseCssAst() here? csstree omit redundant
whitespaces and collapses others to a single space. Its generation is
"round-trip" with parsing (except a single case at the moment, but it's an
edge case: +/*comment*/123):
https://github.com/csstree/csstree/blob/9aa7bee5110b001eb2fa94129d02f0
f385a29bdc/test/translate.js#L15-L28


I did try without it and it typically complained about difference with a
space (or not) after the colon in rule declarations.

In src/browser-sandbox/pruneNonCriticalCss.js
<https://github.com/pocketjoso/penthouse/pull/209#discussion_r154821885>:

>        // check what, if any selectors are found above fold
-      rule.selectors = rule.selectors.filter(isSelectorCritical)
-      return rule.selectors.length > 0
+      const selectors = rule.prelude.value.split(',')

That's incorrect approach, since pseudo classes, attribute selector value
and comments can contain a comma (e.g. :not(a, b) /* comme,nt */
[foo="hello, world"]).
It's better to parse a selector and filter on tree traversal.


Yeah you're right. I wanted to ask you about this: if I let css tree parse
selectors, does it offer a method for generating the selector string I
require, from individual prelude children? So I can skip reimplement your
code for that. :)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/pocketjoso/penthouse/pull/209#pullrequestreview-81039687>,
or mute the thread

src/browser-sandbox/pruneNonCriticalCss.js
pocketjoso: (moving this comment here where it's better placed):

test/util/normaliseCssAst.js
```diff
@@ -1,7 +1,17 @@
...
+    parseRulePrelude: false,
+    parseAtrulePrelude: false,
+    parseValue: false
+  })
+}
 // because dont want to fail tests on white space differences
-export default function normaliseCssAst (cssString) {
-  return css.parse(css.stringify(css.parse(cssString), { compress: true }))
+export default function normaliseCssAst (css) {
+  return parseCssAst(
```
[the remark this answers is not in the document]
pocketjoso: (same moving the comment):

lahmatiy:
> if I let css tree parse selectors, does it offer a method for generating the selector string I
require, from individual prelude children? So I can skip reimplement your
code for that. :)

There is no special method for generating the selector string but `csstree.translate()` works for any node. You should use `context` option to parse a selector and then you can translate it back using `translate()`.

```js
try {
    var ast = csstree.parse(rule.prelude.value, { context: 'selectorList' });
    // do something with ast
    rule.prelude.value = csstree.translate(ast);
} catch(e) {
    // rule.prelude.value can't be parsed as a selector list
}
```

lahmatiy:
>> Did you try to use a single parseCssAst() here? ...
>
> I did try without it and it typically complained about difference with a space (or not) after the colon in rule declarations.

I meant after switching to csstree. For csstree this is probably not necessary.

pocketjoso:
>>> Did you try to use a single parseCssAst() here? ...
>>
>> I did try without it and it typically complained about difference with a space (or not) after the colon in rule declarations.

>I meant after switching to csstree. For csstree this is probably not necessary.

Yep, I meant after too. I'll try it again when I get some time and I'll show you the unwanted diffs.

pocketjoso:
@lahmatiy re `parse` + `context` + `translate` - as long as that's not overly expensive to do for every single selector in a whole css file (I imagine it's not really a problem) then 👍 sounds like just what I was looking for. Will implement and re-push here before merging.

lahmatiy:
@pocketjoso As I see you switch off rule prelude parsing:

```js
csstree.parse(css, {
    parseRulePrelude: false,
    ...
});
```

That's mean csstree consumes a selectors as a balanced raw strings, i.e. with no deep analysis and this saves a time and memory. When you parse rule preludes later, it takes additional time, not so much but the time. Since you turn AST back to string it may lead to GC trashing. Not sure that's a big problem.

Anyway, it may be better to make a detailed parsing CSS in the beginning (at least rule and atrule preludes, since you process it). It helps avoid parse-translate code everywhere (and try-catch blocks since a stylesheet parsing doesn't throw an errors) and GC trashing. In that case wrong selectors (i.e. rule.prelude) and atrule preludes will be presented as a Raw nodes and rules with such prelude can be omitted (filtered).

pocketjoso:
Yeah I was not clear, I meant to turn `parseRulePrelude: true`,  and my question was just about how to turn the parsed rule prelude back to strings - which it sounds like `translate` could be enough for.

lahmatiy:
@pocketjoso Ok. There is no special method to collapse selectors to a single node at the moment. So you need do something like that:

```js
const selector = csstree.translate(rule.prelude);
// in case you need to save a selector as a single node
rule.prelude = {
    type: 'Raw',
    value: selector
};
```

I think it's a good idea to implement expand/collapse methods to address this task. I've created an issue for that https://github.com/csstree/csstree/issues/70

pocketjoso:
My use case is really just to replace the `parseRulePrelude: true` + `rule.prelude.value.split(',')` into `parseRulePrelude: false` and whatever it takes me to get the same array of css selectors from the (parsed) prelude.

pocketjoso:
`rule.prelude: {"type":"SelectorList","loc":null,"children":[{"type":"Selector","loc":null,"children":[{"type":"TypeSelector","loc":null,"name":"body"}]}]}`

@lahmatiy Not getting how to get an array of selectors (same as the `.split(',')` achieved, but in a proper way) via `translate`. If I call `csstree.translate` with either the `rule` or `rule.prelude` from above... I always get `{}` in return. Any advice?

-------

In general here unlike what I said yesterday I will hold of merging this and try to get the proper selector "prelude" parsing in first. Should be later this week.

lahmatiy:
> Not getting how to get an array of selectors (same as the .split(',') achieved, but in a proper way) via translate

No sure why you need an array of selectors as a strings, because it's better to manipulate CSS using AST instead of regexps. Anyway, you can do it by this way:

```js
if (rule.prelude.type === 'SelectorList') {
    const selectors = rule.prelude.children.map(csstree.translate);
    // do something with selectors
} else {
    // type === 'Raw' – this means that the selector was not parsed due to errors
}
```

You also can turn a List into an Array by call `toArray()` method on a List instance.

pocketjoso:
@lahmatiy it just doesn't work for me, I still just get `{}` returned for each Selector:
```js
if (rule.prelude.type === 'SelectorList') {
  console.log('rule.prelude.children: ' + JSON.stringify(rule.prelude.children))
  selectors = rule.prelude.children.map(window.csstreeTranslate) // NOTE: this is the method from css.translate
  console.log('selectors: ' + JSON.stringify(selectors))
}
```

yields:
```
rule.prelude.children: [{"type":"Selector","loc":null,"children":[{"type":"TypeSelector","loc":null,"name":"table"},{"type":"WhiteSpace","loc":null,"value":" "},{"type":"TypeSelector","loc":null,"name":"td"},{"type":"AttributeSelector","loc":null,"name":{"type":"Identifier","loc":null,"name":"class"},"matcher":"*=","value":{"type":"String","loc":null,"value":"\"col-\""},"flags":null}]},{"type":"Selector","loc":null,"children":[{"type":"TypeSelector","loc":null,"name":"table"},{"type":"WhiteSpace","loc":null,"value":" "},{"type":"TypeSelector","loc":null,"name":"th"},{"type":"AttributeSelector","loc":null,"name":{"type":"Identifier","loc":null,"name":"class"},"matcher":"*=","value":{"type":"String","loc":null,"value":"\"col-\""},"flags":null}]}] +1ms
selectors: [{},{}] +1ms
```

Perhaps it is because I'm still here inside a browser environment and not calling the csstree.translate method on csstree..

lahmatiy:
@pocketjoso I think that's because in browser env your AST is a plain objects. Try this:

```js
selectors = csstree.fromPlainObject(rule.prelude).children.map(window.csstreeTranslate)
```

pocketjoso:
Just an update - indeed you were right @lahmatiy and I'm wrapping up the change as you suggested tonight.

src/browser-sandbox/pruneNonCriticalCss.js
```diff
@@ -131,51 +131,121 @@ export default function pruneNonCriticalCss ({
...
+    // check what, if any selectors are found above fold
+    let selectors = []
+    if (rule.prelude.type === 'SelectorList') {
+      selectors = await window.translateSelectorListRuleToStrings(rule)
+    }
+    if (selectors.length === 0) {
+      return false
+    }
+
+    // filter out the ones that are not critical
```
lahmatiy: This part may be simpler:
```js
const criticalSelectors = selectors.filter(isSelectorCritical);
const keepRule = criticalSelectors.length > 0;

if (keepRule) {
    rule.prelude.children = criticalSelectors;
}

return keepRule;
```

pocketjoso: This is my array of selector strings, _not_ the `rule.prelude.children` that are part of the `css-tree` AST - that's why my code is more complicated.

In this PR in general I have mostly kept the existing code structure, but I'm aware it's not the best fit for css-tree. It would be good to explore refactoring to stay closer to css-tree’s api's to see what effect that can have both on performance and on code quality - but that will be for follow up PR's.

pocketjoso:
**Update**: now using `css-tree` more fully, also for parsing the rule `prelude`s (i.e. ~ the selector list), as suggested by @lahmatiy.

I have pushed this updated as `penthouse@1.2.0-rc` (on the `rc` tag), and I'm am doing some more serious tests before merging. This should be merged in the next few days unless some regressions or otherwise unforeseen problems come up.

src/browser-sandbox/pruneNonCriticalCss.js
pocketjoso: @lahmatiy The insight I never gave you before while talking about this is that the selector strings are eventually uses to query the browser DOM (`document.querySelector`), that is why I need them in this format.

lahmatiy: Yep, I see that `selectors` is an array of strings. Since you use `rule.prelude.children.filter()` I suppose children is an array too. So suggested code should works (however I'm not sure about it)

pocketjoso:
I should say too that I suspect that this last update (csstree prelude parsing) will make Penthouse slightly slower (mainly because of the async overhead it brings because of how it has to be implemented for puppeteer usage). I am monitoring this, with the expectation that the gain in accuracy will trump the relative speed loss. There are still plenty of room for performance optimisations in Penthouse, both within the code surrounding `csstree` and elsewhere.

src/browser-sandbox/pruneNonCriticalCss.js
pocketjoso: The code could be cleaner and better commented - I will look into that going forward - here I just want to get this PR into a mergable state as it dragged a bit over time vs what I had originally wanted. :)

`rule.prelude.children` is indeed an array, but not of strings, it’s the parsed csstree ast values. It’s simply an array because I call `csstree.toPlainObject` early on.

Hence here we have one array of csstree prelude children, and one of css selector strings, hence why the code is as it is. Surely can be improved.

pocketjoso:
At a first look from some heavier (parallel) usage, this latest version is quite a regression in terms of performance.

I see roughly doubled CPU usage, and more than twice as slow jobs.

Now for regular consumer usage with limited parallel processing and a limited number of jobs to execute, this probably is not noticeable at all. I have to decide whether to push this version anyway and look for performance gains going forward, or revert to the first release candidate and look for alternative ways rewriting the `split(',')` hack. As hacky as it is, at the end of the day today it affects very few rules in production css on the web as far as I can tell. Correct me if you know better @lahmatiy, but what CSS selector syntax exists that allows `,` usage inside a single selector, that _also_ has support in modern browsers already? You mentioned the `:not(a,b,c)` example before, but that is not supported yet, and not really in use at all from what I can see in my own stats.

In either case, I will try first to find a way to bring in the csstree methods I need into puppeteers JS sandbox _without_ using the `exposeFunction`. I suspect that is the bottle neck. I don't know for sure, but I think this is function in puppeteer acts like a bridge from browser JS land to ... another part of Chrome's code base, which could explain the cost. It’s probably not really meant to be called say 1 million times in a few seconds, while traversing over the rules in a css file. @addyosmani - do you have insight (or colleagues around you ;) ) that can verify whether this is true? (i.e. slow and not recommended to essentially call a function that is available via puppeteer.exposeFunction in the middle of a (long) loop)

lahmatiy:
@pocketjoso I think it's better to inject entire `css-tree` into browser context using `page.addScriptTag()` and use it as in node's context. It should make things simpler.

pocketjoso:
Right, I'll try that too.

I should also say that without looking further it's also possible that the stats I have are skewed due to very slow outliers - individual critical css generation calls that essentially hang. I don't think so but could be the case.

lahmatiy:
> but what CSS selector syntax exists that allows , usage inside a single selector, that also has support in modern browsers already?

`:not(<selector-list>)`, `:matches(<selector-list>)`, `:has(<selector-list>)`, `:nth-[last-]child(<nth> of <selector-list>)` – all supported by Safari at the moment
`[attr="hello, world"]` – comma can be used in attribute value
Comma can be escaped and used in any identifier ;) So this is a correct selector: `#hello\,\ world {}` that matches to element with `id="hello, world"`

pocketjoso:
> :not(<selector-list>), :matches(<selector-list>), :has(<selector-list>), :nth-[last-]child(<nth> of <selector-list>
Yeah, but with only don't think many people are shipping such code into production yet, and that's what I saw reflected in my own data too. So yes, it is valid (in CSS4), it is coming, but currently I believe it accounts for a minimal amount of production css. Meaning that while I want to support this too, I believe the increase in accuracy via adopting css-tree and getting fault tolerant css parsing in itself is much greater than the loss of inaccurate `,` splitting in selectors.

.. I'll see what I can make happen in the next few days, but I am considering merging the first release candidate in the mean time. It runs great, has much better stats than any other version of Penthouse.

pocketjoso:
@lahmatiy I should ask - the work that causes these performance regressions (whether it's coming from css-tree calls, puppeteer exposeFunction or async generator overhead) is just for getting a selector string for each `rule.prelude.children` child. Have you considered adding this by default (or via an option) when parsing the css? This would just magically make my problems here go away, and I'm guessing, but could be wrong, that it would not be very expensive for you. It also seems like most of the other css parsers support it in one way or another.

pocketjoso:
Rewrote the css-tree injection again, avoiding the need to make the core pruneNonCriticalCss pipeline async. Much cleaner code, and so far tests look good - now will test at larger scale to see how this compares performance wise with `0.12.0rc` - prior to css-tree prelude parsing.

(I force pushed and replaced the commits I added yesterday for async handling, no need for them. I re-used quite a lot of the non async part of the code changes)

pocketjoso:
So far performance with this rewrite is on par with the performance for the first release candidate, meaning this now looks promising to be merged! ✨ If nothing new comes up, rough plan will be to merge and publish new Penthouse version in about 24h.

It’s available as `penthouse@1.2.0=rc4` in the mean time.

pocketjoso:
All looking good, merging!

lahmatiy:
> getting a selector string for each rule.prelude.children child. Have you considered adding this by default (or via an option) when parsing the css?

@pocketjoso Currently I don't plan to add this. It was in css-tree in the past, but I discarded it. It a bit hard to explain the reasons, the problem it makes parser complicated and confused in some cases (a Selector can be used inside a functional pseudo and at-rule prelude).
Btw, I suppose you need an array of selectors as a strings, because you have old solutions that stayed untouched at the moment. I believe, when you refactor those solutions you'll no need for selectors as strings. Or that's will be a case to think about your suggestion again ;)

pocketjoso:
@lahmatiy I believe I will always need selectors as string, as that's what the [querySelector api requires](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector).

I understand that it could bring some complexity to css-tree. In the last refactor after eliminating the cost of using `exposeFunction` and `async` overhead I do see a performance penalty for having doing the conversion to selector string for each rule in Penthouse as it’s done now, especially on large stylesheets, but it's OK for now, just letting you know.

PR #211, opened 2017-12-12
[BUG] return value of debuglogger expected to be an error or empty st…

hummal:
…ring. If no empty string debuglogger returns undefined

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -308,11 +308,10 @@ module.exports = function (options, callback) {
   const debuglog = function (msg, isError) {
     if (isError) {
       console.error(msg)
-    } else {
-      debuglogger(msg)
-      msg = ''
+      return msg
     }
-    return msg
+    debuglogger(msg)
+    return ''
   }
 
   const logging = {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -311,6 +311,7 @@ module.exports = function (options, callback) {
       return msg
     }
     debuglogger(msg)
+    return ''
```
pocketjoso: I see, this is an old code path that hasn't really been maintained.

pocketjoso: To make a little bit more sense, please return `msg` instead. Also feel free to do an `if/else` with the `console.error/debuglogger` and put the (shared) `return msg` after the `if/else`. I.e.

hummal: but msg is not allowed to be more than "". If I use msg, there is always content in strErr. I would still need to set msg = ''

pocketjoso: > but msg is not allowed to be more than ""

pocketjoso: Better to have it as you had it and just return `` when no error. Sorry I didn't realise exactly what you were trying to fix before.

PR #217, opened 2017-12-16
All postformatting done on ast

pocketjoso:
improves the rest of the postformatting so that none of it operates on a css string, but instead always on the ast format - in my preliminary tests that should be faster. It could be made even faster by doing all the transformations in a single iteration, instead of once per postformatting process, but that's for the future (if it's even worth it, would have to do some more specific measurements first; normally now these operations are very fast..)

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -1,4 +1,5 @@
 import fs from 'fs'
+import csstree from 'css-tree'
 import debug from 'debug'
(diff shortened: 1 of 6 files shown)


pocketjoso:
I'm seeing some regressions with the font-face removal updates, at least in how they relate the the before/after screenshot generation. Will hold of merging this until I've fixed this.

pocketjoso:
Fixed now, merging.

PR #222, opened 2018-01-16
Refactoring of using css-tree

lahmatiy:
I reworked code a little bit to use AST on each stage and use power of CSSTree to manipulate it. Also I changed tests to reduce unnecessary code and fix formatting.


pocketjoso:
Thanks for the PR! sorry for slow response, have been busy.

Overall looks good, will make specific comments in PR now. However I have a concern that the changes you've made will increase execution times. I am thinking I will publish a test version on npm and start using it in my projects to get some numbers.. if you already ~ benched these changes please let me know what you found.

pocketjoso:
Did some quick manual not overly reliable benchmarking just via the test suite, and this update only seems to have made Penthouse a little bit slower. Ideally this would be fixed before merging, but let's see what's feasible. From my quick tests, here's what seems to be the cause of the slowdown:

1. initiating the sandboxed `pruneNonCriticalCss` execution (not the _execution_ of the function) This is the bigger slowdown that I think we should be able to address before merging. Penthouse now takes at least 100ms longer in each simple test, and it looks like the slow down correlates with the original css size, which is kind of bad news (test css files are quite small). I think the arguments to a sandboxed puppeteer function like this one goes through some kind of serialisation, and I would guess this is what has gotten slowed down. Why:
* I imagine that removing `parseAtrulePrelude: false, parseValue: false` increased the size of the `ast` object. Further, it's possible that it is is bigger as an `object` than the array version that `csstree.toPlainObject` returned.

Since it's not really execution time of our own code that is slower here, but rather a serialisation cost we incur, I wonder if this can be bypassed by moving the `csstree.parse` call _into_ the sandboxed `pruneNonCriticalCss` function? Assuming that the original css string is faster to serialize, this could avoid increase in timing. I would necessitate moving the `nonMatchingMediaQueryRemover` call into that function as well, which I think would make sense anyway.

----

Note: I'm listing two more items below, but my conclusion for them is that we can leave them for future optimisations.

2. `processCssRules` was more than twice as slow, but I tried what I commented on in the code - re-ordering so you prune `Atrules` before `Rules`, and now it's only a bit slower. Don't think we can do much better than that, so should be OK for now after that change...

3. `postformatting` (and preformatting) is several times slower, presumably because walking the critical css is not incredibly cheap, and it's now done many times. I think this as well is OK for now. What I want to try in the future is to only walk the critical css _once_ for all postformatting.. but that's a bigger code re-write.

---

So all in all, @lahmatiy: Thanks again for the PR, overall I really like your changes. Perhaps it's not obvious from the outside that I care a lot about execution time for Penthouse jobs, if so my code review might be hitting you as a surprise. I might have told you a bit about this before, but if not: the reasons is that myself and others use Penthouse to process never ending queues of jobs for critical css generation, and thus execution time really matters.

Do you think you will be able to address point `1.` here before merging? Worst case I have to look at it afterwards otherwise, it will just take longer before this update would be released if so.

I'll give my code review comments now as well.

src/browser-sandbox/pruneNonCriticalCss.js
```diff
@@ -132,103 +135,78 @@ export default function pruneNonCriticalCss ({
...
-    if (
-      rule.type === 'Atrule' &&
-      (rule.name === 'charset' ||
-        rule.name === 'import' ||
-        rule.name === 'namespace')
-    ) {
-      return true
-    }
+    csstree.walk(ast, {
+      visit: 'Atrule',
```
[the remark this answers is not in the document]
pocketjoso: Verified that it slows Penthouse down doing it in this order, so please inverse it.

src/core.js
```diff
@@ -35,7 +35,7 @@ async function blockJsRequests (page) {
 async function pruneNonCriticalCssLauncher ({
   browser,
   url,
-  astRules,
+  ast,
```
pocketjoso: regarding my performance comments - this is the relevant line (for 1.) - passing this (bigger) ast object here slows down the initialisation of this function, probably during parameter serialisation. Being able to pass through something that is quicker to serialize (smaller) should resolve this (particular) performance regression.

src/core.js
```diff
@@ -174,25 +174,32 @@ async function pruneNonCriticalCssLauncher ({
         debuglog('take before screenshot DONE: ' + beforePath)
       }
 
-      const astRulesCritical = await page.evaluate(pruneNonCriticalCss, {
-        astRules,
+      // remove all comments
+      csstree.walk(ast, {
```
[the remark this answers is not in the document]
pocketjoso: Given my (quick!) performance analysis, I also think this slows down code execution (did not yet benchmark this line in isolation however). For this reason, I would remove this and leave it to a dedicate minifier to do this job.

src/postformatting/unused-fontface-remover.js
```diff
@@ -1,59 +1,77 @@
...
       }
-      if (font) {
-        toKeep = fontNameValues.some(value => value.indexOf(font) !== -1)
-        if (!toKeep) {
-          debuglog('drop unused font-face rule: ' + font)
+
+      csstree.walk(atrule, {
+        visit: 'Declaration',
+        enter: declaration => {
+          if (csstree.property(declaration.property).name === 'font-family') {
```
pocketjoso: Add your new font-face check here instead so you don't have to `walk` one extra time for it (if no `src`, remove).

src/postformatting/embedded-base64-remover.js
```diff
@@ -1,60 +1,61 @@
...
-      })
-      .filter(Boolean)
+  })
 
-    return rule
-  }
-  return rule
-}
+  // remove @font-face atrules with no src declaration
+  csstree.walk(ast, {
```
pocketjoso: unnecessary to do another full `walk` on the whole ast for this - do it inside the `unused fontface remover` instead, I made a comment about it. Cheap since you area already iterating over the font-face rules.

src/postformatting/unwanted-properties-remover.js
```diff
@@ -1,41 +1,44 @@
...
+    visit: 'Declaration',
+    enter: (declaration, item, list) => {
+      if (shouldRemove(declaration)) {
+        list.remove(item)
+      }
+    }
+  })
+
+  // remove empty rules
+  csstree.walk(ast, {
```
pocketjoso: Hmm this is a bit slow in my tests. Not sure we can do it any faster though, let me know if you have ideas.

src/postformatting/unwanted-properties-remover.js
```diff
@@ -1,41 +1,44 @@
...
+    visit: 'Rule',
+    leave: (rule, item, list) => {
+      if (rule.block.children.isEmpty()) {
+        list.remove(item)
+      }
+    }
+  })
+
+  // remove empty at-rules
+  csstree.walk(ast, {
```
pocketjoso: Why is this needed given the check you added [here](https://github.com/pocketjoso/penthouse/pull/222/files#diff-71b8f79184f85a105511f48046105fe2R165)? Perhaps because postformatting could have emptied the rule.. Hmm, ok with this though, not as expensive as walking every rule in the ast.

src/postformatting/unused-fontface-remover.js
```diff
@@ -1,59 +1,77 @@
...
+    }
+  })
+
   return fontNameValues
 }
 
-function unusedFontfaceRemover (rules) {
+function unusedFontfaceRemover (ast) {
   debuglog('getAllFontNameValues')
   // NOTE: we grabp the full declaration everywhere a font name is used,
```
pocketjoso: this comment is not relevant anymore, please remove/update.

src/browser-sandbox/pruneNonCriticalCss.js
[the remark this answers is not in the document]
lahmatiy: I rely on current test suite, so I tried repeat the logic. As I see now, proposed solution is a little bit different from original one. Not sure we can place it first, because `@media`, `@document` and `@supports` are testing for emptiness. I think at-rule filtering based on name should be in pre-processing and test for empty blocks should be at the ending of post-processing, both outside of sandbox.

src/core.js
lahmatiy: Yes, a tree become bigger. I think we should produce a selector list from AST (array of strings) and pass that list to Puppetteer to filter it. It avoid us from passing a big JSON between processes and evaluating CSSTree into browser context (CSSTree doesn't require in browser context, because all work with AST can be done outside of it).

src/core.js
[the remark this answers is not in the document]
lahmatiy: Well, without this check a some tests fails. Original solution filter comments on top level by incident. We can limit this filtering to top level only, like original solution does.

src/postformatting/embedded-base64-remover.js
lahmatiy: Agree, it can be moved it to font-face remover. But we need swap the order of ffRemover and embeddedbase64Remover, and change related tests, since embeddedbase64Remover tests check at-rules with no src is removed. If that's ok, I'll change it.

src/browser-sandbox/pruneNonCriticalCss.js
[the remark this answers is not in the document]
pocketjoso: until at-rule filtering lives in preprocessing (which based on my other comments might end up inside this function anyway), we should just reverse these two to prevent unnecessary processing. The order used previously did not matter as the css was traversed in chronological order.

src/browser-sandbox/pruneNonCriticalCss.js
```diff
@@ -132,103 +135,78 @@ export default function pruneNonCriticalCss ({
...
-    ) {
-      return true
-    }
+        /* Case 1: @-rule with CSS properties inside [REMAIN]
+           @font-face, @keyframes - keep here, but remove later in code, unless it is used.
+        */
+        if (
+          name === 'font-face' ||
+          name === 'keyframes' ||
+          name === 'viewport'
```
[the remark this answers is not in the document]
pocketjoso: Great 👍

src/core.js
pocketjoso: Good idea, that would improve performance even further! Are you interested in taking on such work, or should we look to close this PR first, and look at that work separately?

lahmatiy:
@pocketjoso Thank you for review. First of all I tried to use CSSTree in a better way and try its new API on your tasks. I don't care so much about a performance, but like you, I want make it fast as possible and I see many ways to improve Penthouse and CSSTree. However, I spent more time on the PR than planned, and I needed your feedback to understand that I'm on the right track. Now I can continue to work on it. Will update PR soon.

src/core.js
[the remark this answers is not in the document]
pocketjoso: I'm happy with either of:

src/postformatting/embedded-base64-remover.js
pocketjoso: Go for it 👍

pocketjoso:
@lahmatiy you are awesome, keep it up!

For keeping an eye on performance, for now `env DEBUG='penthouse, penthouse:*' yarn test` is your best friend, but it is quite limited, as only the first test contains a somewhat more representative stylesheet size wise.

I think it would be useful to create a new `yarn test--perf` that just runs penthouse on a bunch of larger stylesheets basically just to get out some execution times, this way we can keep an eye out on performance changes more easily. Feel free to add to your PR (should be quite easy), or let me know and I can push it to master (but will most likely not have time until next weekend).

lahmatiy:
I guess I'm done on PR. Last changes:
- pass selector list to sandbox only (instead of entire AST) to filter non-critical selectors
- remove CSSTree injection into sandbox as not needed anymore
- rework removers to split responsibility and reduce traversal count
- PR fixes

Performance should be better now. For `should match exactly the css in the yeoman test` test time is down from 1400-1500ms to 900-1000ms on my machine. Time split:
- `browser launch` + `new page` + `viewport set` – takes ~500ms
- `page load` + `wait for renderWaitTime` – ~230ms
- `css parsing` – 120-180 ms; it's quite surprising to me, since when this parsing is running in a separate script it takes just ~60ms. Have no clue what the reason at the moment
- the rest work, i.e. prune non-critical selectors and AST cleanup, takes ~100ms

pocketjoso:
First of all - 👏 this is great work for Penthouse! Execution times are now lower than before this PR, which is great to see.
The question(s) you raised (f.e. why `css parsing` takes longer than when run in separate script), and other improvements (make sure we run as much as possible in parallel) can happen later. I will look at the code again now, and hopefully merge tonight or tomorrow!

src/browser-sandbox/pruneNonCriticalSelectors.js
```diff
@@ -0,0 +1,101 @@
...
+    // we have a selector to test, first grab any matching elements
+    var elements
+    try {
+      elements = document.querySelectorAll(selector)
+    } catch (e) {
+      // not a valid selector, remove it.
+      return false
+    }
+
+    // filter a selector if all elements are above the fold
```
pocketjoso: this comment is a bit off. The selector should be filter out if none of the matching elements are above the fold, otherwise it should be kept. The code looks right.

src/postformatting/rule-selector-remover.js
```diff
@@ -0,0 +1,33 @@
...
+        (selectorNode, item, list) => {
+          let decision = selectorNodeMap.get(selectorNode)
+          if (typeof decision === 'string') {
+            decision = selectors.has(decision)
+          }
+          return typeof decision !== 'boolean' || decision
+        }
+      )
+
+      // NOTE: isCssSelectorRuleCritical mutates the rule
```
pocketjoso: This comment is in the wrong place now, I think it can just be removed?

src/selectors-profile.js
```diff
@@ -0,0 +1,110 @@
...
+// detect these selectors regardless of whether one or two semi-colons are used
+var pseudoSelectorsToKeepRegex = pseudoSelectorsToKeep
+  .map(function (s) {
+    return ':?' + s
+  })
+  .join('|') // separate in regular expression
+// we will replace all instances of these pseudo selectors; hence global flag
+var PSUEDO_SELECTOR_REGEXP = new RegExp(pseudoSelectorsToKeepRegex, 'g')
+
+// TODO: bind with forceInclude argument instead
```
pocketjoso: comment irrelevant now (had to do with execution in the sandboxed browser environment previously)

src/postformatting/index.js
```diff
@@ -1,12 +1,11 @@
-import debug from 'debug'
-
 import commentRemover from './comment-remover'
 import embeddedbase64Remover from './embedded-base64-remover'
 import unusedKeyframeRemover from './unused-keyframe-remover'
 import unusedFontFaceRemover from './unused-fontface-remover'
 import unwantedPropertiesRemover from './unwanted-properties-remover'
 import ruleSelectorRemover from './rule-selector-remover'
 import finalRuleRemover from './final-rule-remover'
+import debug from 'debug'
```
pocketjoso: not so important but I prefer external imports at the top of the file.

src/postformatting/index.js
```diff
@@ -43,6 +42,4 @@ export default function cleanup ({
   // remove empty and unwanted rules and at-rules
   finalRuleRemover(ast)
   debuglog('finalRuleRemover')
-
-  return ast
```
pocketjoso: so the ast is modified (mutated) and hence does not need to be returned; is that it?

src/postformatting/rule-selector-remover.js
```diff
@@ -24,7 +23,7 @@ export default function ruleSelectorRemover (ast, selectorNodeMap, selectors) {
         }
       )
 
-      // NOTE: isCssSelectorRuleCritical mutates the rule
+      // remote the rule if no selector is left
```
pocketjoso: typo: remove

src/postformatting/rule-selector-remover.js
```diff
@@ -0,0 +1,32 @@
+import csstree from 'css-tree'
+
+export default function ruleSelectorRemover (ast, selectorNodeMap, selectors) {
+  selectors = new Set(selectors)
```
pocketjoso: This is ace - for large stylesheets this saves considerable bytes on the data needed to pass to the sandboxed function (to serialise) and hence makes Penthouse significantly faster! Will test in larger scale and come back with some numbers, but really excited about this! 💯

pocketjoso: I had noted before that just initialising the sandboxed function could take significant time for larger stylesheets but I never got down to exactly why and what could be done about this. So happy to see this happening now. :)

src/postformatting/rule-selector-remover.js
lahmatiy: Hm, looks like you comment an outdated source, because I remember that I deleted this comment

src/selectors-profile.js
lahmatiy: The same as above, I deleted this comment

src/postformatting/rule-selector-remover.js
lahmatiy: It was fixed in this commit https://github.com/pocketjoso/penthouse/pull/222/commits/aaaf459e329076fe495cd90e23dcf23a4718ff9c

src/postformatting/index.js
lahmatiy: Yep, that's right

src/postformatting/index.js
lahmatiy: I didn't know the logic of import order. I decided to keep the same order across modules, in most modules it was last one.

src/postformatting/rule-selector-remover.js
lahmatiy: It would be nice to know results of your benchmarks ;)

lahmatiy:
Thank you for your soon review and merge! 🎉

src/postformatting/rule-selector-remover.js
pocketjoso: Yep indeed, I looked at the commits one by one - and I already fixed them in master too, so please just ignore (the ones about changes). All good! 👍

src/postformatting/rule-selector-remover.js
pocketjoso: So about 65% faster compared to previous version, as I shared in my tweet. It also seems to have resolved a small number of issues with generated criticalcss, presumably because of improved selector parsing. 👍

pocketjoso:
@lahmatiy Just FYI I'm working on a follow up PR to do the css parsing and preformatting in parallel with the page loading, which will further speed up Penthouse, inspired by the gains from this PR. :)

pocketjoso:
The update from this PR is now released in `penthouse@1.4.0`

PR #223, opened 2018-01-24
tests: add a simple performance test for small collection of html pages

pocketjoso:
note: these pages have external assets (images, video), whose load time will affect test,
so ideally would replace those with locally saved copies.

Just to make it a bit easier to keep track of execution timings for Penthouse in more real life conditions (primarily: larger stylesheets).


pocketjoso:
cc @lahmatiy - a simple start.

PR #224, opened 2018-01-24
Further perf improvements

pocketjoso:
Following #222, this PR parallelize some of the more expensive operations in penthouse, reducing execution time further. Will run some larger tests in my projects and report on average savings vs master.

### DIFF
--- a/src/browser-sandbox/pruneNonCriticalSelectors.js
+++ b/src/browser-sandbox/pruneNonCriticalSelectors.js
@@ -4,7 +4,7 @@ export default function pruneNonCriticalSelectors ({
   selectors,
   renderWaitTime
 }) {
-  console.log('debug: pruneNonCriticalSelectors')
+  console.log('debug: pruneNonCriticalSelectors init')
   var h = window.innerHeight
 
   // cache whether elements are above fold,
--- a/src/core.js
+++ b/src/core.js
@@ -5,6 +5,7 @@ import pruneNonCriticalSelectors
 import replacePageCss from './browser-sandbox/replacePageCss'
 import cleanupAst from './postformatting'
 import buildSelectorProfile from './selectors-profile'
+import nonMatchingMediaQueryRemover from './non-matching-media-query-remover'
 
 const debuglog = debug('penthouse:core')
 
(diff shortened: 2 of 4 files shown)


src/non-matching-media-query-remover.js
```diff
@@ -69,7 +69,11 @@ function nonMatchingMediaQueryRemover (
       if (csstree.keyword(atrule.name).name !== 'media') {
         return
       }
-
+      // this can happen - why? (atrule.prelude === null)
+      // and should we remove this rule here, or keep it?
+      if (!atrule.prelude) {
```
pocketjoso: @lahmatiy This seems to come from #222 - came across this error in production, have not yet mapped it to original css, but the prelude was `null`, and `children` was `[]`.. anything you can say already just based on that info? (should this check be needed etc)

pocketjoso:
Not a clear winner in performance over previous version, so will keep testing it to investigate what can be improved.

src/non-matching-media-query-remover.js
lahmatiy: This may be the case if there is an error in CSS, e.g. `@media { }` – no prelude here and children is empty. Since CSSTree parser is tolerant to errors it doesn't validate a CSS for correctness of specific at-rules or properties, it just make a check of basic structures (i.e. at-rule can contain a prelude and a block, but both are optional). Later you can use lexer to validate AST against the CSS rules. At the moment lexer can validate a declarations, support for other parts of CSS (like at-rules) will be added in next releases and it will allow to filter a bad nodes before the rest processing.
You can set [`positions: true`](https://github.com/csstree/csstree/blob/master/docs/parsing.md#positions) option to `parse()` and look at `loc` field of a specific node to find a source of problem.

pocketjoso:
Didn't change this PR yet but after some debugging I realised that my instrumentation for comparison penthouse versions (not part of this repo nor open source) needed improvements. After doing that work I can now see that on average this version _is_ faster than `master`, but there are some cases, for some domains, where it is slower. So I'm leaning towards merging and then continuing looking at the cases where it's slower to try to avoid regressing the execution time for these domains.

Issue #202, opened 2017-11-20
Avoid page navigation during critical css generation (causes errors)


hummal:
For some URLs Iam checking Iam getting an strange error, which is properbly puppeteer releated. Think of updating to a newer version.

Message was:
`{ Error: Error: Protocol error (Runtime.callFunctionOn): Execution context was destroyed. undefined
    at /var/www/bamboo_deploy/1190/vendor/tcms/atf-optimizer/node_modules/penthouse/lib/index.js:269:23
    at Generator.next (<anonymous>)
    at step (/var/www/bamboo_deploy/1190/vendor/tcms/atf-optimizer/node_modules/penthouse/lib/index.js:45:191)
    at /var/www/bamboo_deploy/1190/vendor/tcms/atf-optimizer/node_modules/penthouse/lib/index.js:45:361
    at process._tickCallback (internal/process/next_tick.js:109:7)
  stderr: 'Error: Protocol error (Runtime.callFunctionOn): Execution context was destroyed. undefined' }
`

Maybe this is related: https://www.bountysource.com/issues/51332369-when-would-an-error-cannot-find-context-with-specified-id-undefined-happen

Also I want to mention that we need to install a lot of dependeciens with puppeteer. So I will have 3 new users in my passwd file. Keeps me feeling bad.

The problem with the failing id is much more important for me.

pocketjoso:
Based on that discussion, it should be fixed in the next puppeteer release (0.14) - so let's retry then.

pocketjoso:
I know puppeteer install can be quite painful, hopefully that will be improved in the library itself. If you really want to you could revert to `0.11.13`, the version that still used phantomjs. It's no longer maintained however.

hummal:
I hope so. There is a little difference in my error message and the one in the thread. Its about execution context was destroyed and not an id is missing. 

Returning to an older version is not an option. Iam not into using software without updates. :)

hummal:
@pocketjoso what do you expect as output from debuglog function?

Code like this: 
```javascript
 stdErr += debuglog('call generateCriticalCssWrapped');
```
results in _stdErr = undefinedundefinedundefined_

hummal:
Ok I solved the problem by simply adding a return ""; to the end of the debuglog function.
PR incoming

hummal:
@pocketjoso I opend an issue at puppeteer and they had some interesting information. The result would be that you should include these handlings because it could happen to anyone.

[...]

PR #227, opened 2018-02-07
Puppeteer page error - Fix bundle

hummal:
This PR is trying to fix 2 main problems:
- `Cannot find context with specified id undefined`
- `Execution context was destroyed. undefined`

Additional it is taking care of some other minor issues:
✅ added missing dependecies for development
✅ added `.gitattributes` file to ensure correct line endings and file handlings on all os
✅ fixed `package.json` target nodejs version. It has to be string instead of float
✅ moved the `pageLoadSkipTimeout `calculation into the browser **after** the first response (_document_) to ensure a better timing and remove redundancy with timeout
✅ `pageLoadSkipTimeout `now prevents the page to load additional resources after timing out with `window.stop()`
✅ fixed error handling and promise rejecting. Now all pages which throw errors will repeated once like they should and dont break penthouse
✅ handling of https://github.com/GoogleChrome/puppeteer/issues/1591 bug with` page.waitFor(n)`
✅ slightly reformating

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ yarn add --dev penthouse
 ```js
 penthouse({
   url: 'http://google.com',
-  cssString: 'body { color; red }'
+  cssString: 'body { color: red }'
 })
 .then(criticalCss => {
   // use the critical css
--- a/examples/basic.js
+++ b/examples/basic.js
@@ -3,7 +3,7 @@ import fs from 'fs'
 
 penthouse({
   url: 'http://google.com',
-  cssString: 'body { color; red }'
+  cssString: 'body { color: red }'
 })
 .then(criticalCss => {
   // use the critical css
--- a/examples/custom-browser.js
+++ b/examples/custom-browser.js
@@ -13,7 +13,7 @@ const browserPromise = puppeteer.launch({
 })
 penthouse({
   url: 'http://google.com',
-  cssString: 'body { color; red }',
+  cssString: 'body { color: red }',
   puppeteer: {
     getBrowser: () => browserPromise
   }
--- a/examples/screenshots.js
+++ b/examples/screenshots.js
@@ -10,7 +10,7 @@ import penthouse from 'penthouse'
 // 'homepage-after.jpg'
 penthouse({
   url: 'http://google.com',
-  cssString: 'body { color; red }',
+  cssString: 'body { color: red }',
   screenshots: {
     basePath: 'homepage', // absolute or relative; excluding file extension
     type: 'jpeg', // jpeg or png, png default


hummal:
Accoding the failing build: 
https://github.com/babel/babel/tree/master/packages/babel-preset-env

> Similarly, if you're targeting Node.js instead of the browser, you can configure @babel/preset-env to only include the polyfills and transforms necessary for a particular version:
> 
> ```
> {
>   "presets": [
>     ["@babel/preset-env", {
>       "targets": {
>         "node": "6.10"
>       }
>     }]
>   ]
> }
> ```

hummal:
by the way is there a reason why you dont update your dependecies? 
```
 jsesc                    ^1.0.0  →          ^2.5.1
 puppeteer                  0.12  →             1.0
 chai                     ^1.9.1  →          ^4.1.2
 eslint-config-standard  ^10.2.1  →  ^11.0.0-beta.0
 eslint-plugin-node       ^5.2.1  →          ^6.0.0
 husky                   ^0.13.3  →         ^0.14.3
 lint-staged              ^3.4.1  →          ^6.1.0
 mocha                   ^1.20.1  →          ^5.0.0
 prettier-standard        ^5.0.0  →          ^8.0.0

The following dependencies are satisfied by their declared version range, but the installed versions are behind. You can install the latest versions without modifying your package file by using npm update.
 If you want to update the dependencies in your package file anyway, run ncu -a.

 jest      ^22.1.4  →  ^22.2.0
 eslint    ^4.11.0  →  ^4.17.0
 gm        ^1.21.1  →  ^1.23.1
 prettier   ^1.3.1  →  ^1.10.2
```

hummal:
I updated the dependecies to the latest versions (except puppeteer). Removed mocha and chai (because you are using jest) and it worked till the extra and perf tests. It has somehow a problem with creating the penthouse instance.

You should propably check that. I won't commit these changes right now but you can see the results in here: https://travis-ci.org/hummal/penthouse

pocketjoso:
So thanks for the PR, I started reviewing now. There's some stuff in here that is just obvious improvement, but then there are other things that we need to talk about before merging.
I looked at whether we could cherry-pick out the easy changes from your commits, but they were spread out across commits, so [I opened up a separate PR with just those changes](https://github.com/pocketjoso/penthouse/pull/228). If you are up for it, please re-open your own PR with just those changes, so the commits get credited to you, since the changes are yours. Otherwise I will merge that, and we will just continue with the other changes here.

src/browser-sandbox/pruneNonCriticalSelectors.js
```diff
@@ -83,19 +83,25 @@ export default function pruneNonCriticalSelectors ({
...
 
-  // give some time (renderWaitTime) for sites like facebook that build their page dynamically,
-  // otherwise we can miss some selectors (and therefor rules)
-  // --tradeoff here: if site is too slow with dynamic content,
-  // it doesn't deserve to be in critical path.
-  return sleep(renderWaitTime).then(() => {
-    console.log('debug: waited for renderWaitTime: ' + renderWaitTime)
-    return filterSelectors(selectors)
-  })
+  try {
```
pocketjoso: Did you test putting this try/catch inside core.js instead?  As long as that still catches the error, I prefer it, as we can react to the error, instead of just receiving no selectors from this function. I implemented this in [my branch here](https://github.com/pocketjoso/penthouse/pull/228/files#diff-7eb52b366866677666470e019283c8eaR177), but I don't have an easy way to test the issues you have been testing, so I don't know 100% that it catches your errors, but it looks like it should.

src/core.js
```diff
@@ -101,52 +124,105 @@ async function pruneNonCriticalCssLauncher ({
...
-                    's. '
-                )
-                resolve()
-              }
-            }, pageLoadSkipTimeout)
+      const loadPageResponse = new Promise((resolve, reject) => {
+        page
+          .goto(url, {
+            timeout: timeout + 1000,
+            waitUntil: 'networkidle'
```
pocketjoso: Hmm, I tried this value before but it leads some urls to get stuck, if they continuously send requests for tracking and similar. I'm not sure it is a better default for Penthouse. What was your reasoning behind this change?

src/core.js
```diff
@@ -101,52 +124,105 @@ async function pruneNonCriticalCssLauncher ({
...
+          .catch(err => {
+            _hasError = err
+            // Reject when error because we don't want an errored page
+            console.error(err)
+            return reject(err)
+          })
+      })
+
+      try {
+        // Instead of race we want to use all for better workflow. Race is evil
```
pocketjoso: this comment doesn't match the code, did you change your mind or is it a mistake? I am interested in why you think 'Race is evil'? To the best of my knowledge so far, Promise.race is a good fit here.

src/index.js
```diff
@@ -52,7 +53,8 @@ const launchBrowserIfNeeded = async function ({ getBrowser }) {
         // for spammy pages that keep on sending (non critcal) requests
         waitUntil: 'networkidle2',
         ignoreHTTPSErrors: true,
-        args: ['--disable-setuid-sandbox', '--no-sandbox']
+        args: ['--disable-setuid-sandbox', '--no-sandbox'],
+        dumpio: false
```
pocketjoso: This seems to be the default in the [latest version of Puppeteer](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions). Why did you add it; is it not the default on the 0.12 version?

src/core.js
```diff
@@ -166,11 +242,11 @@ async function pruneNonCriticalCssLauncher ({
       )
       debuglog('build selector profile')
 
+      await page.waitFor(500)
```
pocketjoso: this is the kind of change that I most like would want as an opt-in, or even just configurable so that only users who need to change this can do so, without slowing down penthouse for everyone.

src/browser-sandbox/pageLoadSkipTimeout.js
```diff
@@ -0,0 +1,33 @@
...
+  try {
+    return new Promise((resolve, reject) => {
+      if (pageLoadSkipTimeout) {
+        const start = Date.now()
+        const pageLoadTimeout = function () {
+          window.requestAnimationFrame(() => {
+            const timePassed = Date.now() - start
+
+            if (timePassed >= pageLoadSkipTimeout) {
+              window.stop()
```
pocketjoso: The intention with the Promise.race for the page load in `master` is just to prevent Penthouse to take too long, when the page has reasonably loaded, but due to imperfect detection mechanisms puppeteer still believes it has not.

src/core.js
```diff
@@ -101,52 +124,105 @@ async function pruneNonCriticalCssLauncher ({
...
+
+      // Handle requests of pages and block the page from going away and losing context
+      // https://github.com/pocketjoso/penthouse/issues/202
+      let pageLoadSkipPromise = new Promise((resolve, reject) => {
+        if (pageLoadSkipTimeout) {
+          // enables us to control requests(stop, continue, modify)
+          page.on('response', async response => {
+            if (response.url === url) {
+              debuglog('RESPONSE URL: ' + response.url) // TODO: interceptedRequest.url is a function in puppeteer > 0.12
+              if (pageLoadSkipTimeout) {
```
pocketjoso: you already checked this

src/core.js
```diff
@@ -101,52 +124,105 @@ async function pruneNonCriticalCssLauncher ({
...
-              }
-            }, pageLoadSkipTimeout)
+      const loadPageResponse = new Promise((resolve, reject) => {
+        page
+          .goto(url, {
+            timeout: timeout + 1000,
+            waitUntil: 'networkidle'
+          })
+          .then(response => {
+            // Reject exits the Promise.all call immediately and ensures that the pageload is always higher prio than timout
```
pocketjoso: I think you are mixing up Promise.all and Promise.race here - this comment is not accurate. (this fulfills the promise, hopefully winning the race. Would not do what you want if it was promise.all below).

package.json
```diff
@@ -83,7 +85,7 @@
         "env",
         {
           "targets": {
-            "node": 6.4
+            "node": "6.4"
```
pocketjoso: I remember thinking so to, but see the errors  this gives:

src/browser-sandbox/pruneNonCriticalSelectors.js
hummal: The reason for me to do it that way was to ensure to solve the execution context problem. But I tried a lot of things. So I will check if its running safe with the try catch in core. In common you are right.

src/core.js
hummal: This is supposed to your other questions. 
I did the window.stop() in the executed page tab because I want the load event get fired. In this case the timeout wins the race with a correct "window load event". I did some testing with that but after all the changes I made I wouldnt bet it wouldn't run without it.

src/core.js
hummal: The comment is just wrong. I changed it to Promise.all to be more safe. To explain:
The evil thing on race is, if you reject a promise the race is over and the promise race is rejected at all. But what we want to achieve is that the first promise resolve will win. Nevertheless I needed to change the workflow of promise.all to be reversed (reject is win). So I decided it is also not the best way. I just changed my mind. yeah

src/index.js
hummal: I missed to remove this. Just used it to dump all console output while searching for the execution context problem. Can be removed

src/core.js
hummal: Maybe you are right. I need to use it in a very large scale and Iam in need of a working system. If there is a chance of an error my results will be wrong and this is crucial and could cost thousands of Euros.
But of course, we could make an options out of it. Good idea

src/browser-sandbox/pageLoadSkipTimeout.js
hummal: The mindset of stop waiting for the page to load but let it load while extracting the css is helping me to understand. 
As mentioned above I guess it is triggering the "load" event which results in a resolving of the promise. Which is what I want to achieve.

src/core.js
hummal: you are right

src/core.js
hummal: It is a relic of my try to replace race with all. Just a wrong comment.

package.json
hummal: I know that travis is dealing so. But I really dont get why. This is more a travis problem then a bug. Because as mentioned in the documenation it is recommended to use a string. 
What do you want to do now?

hummal:
At all, Iam for vacation till wednesday. I can't pull a new pr until then. If you can't wait, you should go.

pocketjoso:
Okay, so I think what I'm going to do:
1. merge #224 - which I realised after looking at it again not only improves performance of penthouse further (5-10% faster execution time on average), but also cleans up the code a bit.
2. then I will rebase #228 with your ease changes and get that merged as well.
(both of these before Wednesday)

After that we would need to rebase this PR to take up the changes - I would however like you @hummal to give what I wrote in https://github.com/pocketjoso/penthouse/issues/202#issuecomment-364698843 a go first.

And finally, please open a separate PR for your branch with dependency bumps (https://travis-ci.org/hummal/penthouse) - as long as the tests pass (I disabled a performance test that was failing before) we can get that merged quickly too.

Cheers, keep up the good work! 👏

src/browser-sandbox/pruneNonCriticalSelectors.js
pocketjoso: tested and merged this in (via core.js) in #228

src/core.js
pocketjoso: I actually see that Penthouse uses `networkidle2`, but specified in the puppeteer.launch call [here](https://github.com/pocketjoso/penthouse/blob/master/src/index.js#L51)... which according to the existing documentation seems wrong - although they might have moved it in a more recent version.
So all in all, I would be in favour of testing the `networkidle2` value here - but unless it is related to your changes I suggest for you _not_ to do it (in this PR), and I can isolate and test just that change myself (later).

pocketjoso:
So, an update from me:
I did merge #224, rebase the simplest changes you had and merge those as well via #228, and I am currently sanity testing these changes under `penthouse@1.4.1-rc3` - only expecting slightly faster execution on average.

So what remains here? For me the next step would be for you @hummal to try https://github.com/pocketjoso/penthouse/issues/202#issuecomment-364698843 as I said above (ideally on updated master), then if you still have problems, let's continue with the solutions for them in this PR, which would then need to be rebased.

hummal:
Ah I see my comment wasn't submitted before. 
I tried to use puppeteer as browser before, but I can't get it to work. I always run into the error "ERROR: kill ERSCH" and I can't fix it. I remembered that I tried that before. 

As for all the pr requests. I don't have so many time that I want to use it to have my name in a commit I did. If you want to use this extra pr to split the commits, it is ok for me if you just merge yours with my changes. We both know that I supported you. And the software is better, so WIN-WIN.

As I said, Iam not able to test your suggestions of https://github.com/pocketjoso/penthouse/issues/202#issuecomment-364698843 

What else is left?

hummal:
Merged current master into my pr

hummal:
Split into other PRs.

Issue #226, opened 2018-01-29
puppeteer@1.0 upgrade


hummal:
Test & update puppeteer to the released 1.0 version

pocketjoso:
Thanks. Would love to upgrade but is blocked by:
https://github.com/GoogleChrome/puppeteer/issues/1506

which in turn is blocked by:
https://bugs.chromium.org/p/chromium/issues/detail?id=781133#c4

simllll:
Any news on this? 1.1 is out, is it still a blocker? I'm on aws Linux, have no chance to install several actually unsupported versions of puppeteer. I have 1.1 running, will test as soon as I have time. Thought maybe someone has news on this already?

simllll:
My results so far with "my special AWS setup":
I have installed: Google Chrome 64.0.3282.167 

I run penthouse with a custom puppeteer instance and a config like this:
```
penthouse({
                                        url: `file:///${tmpPath}`,
                                        cssString,
                                        forceInclude: [
                                            /^\.sb-/,
                                        ],
                                        puppeteer: {
                                            getBrowser: () => puppeteer.launch({
            executablePath: '/opt/google/chrome/google-chrome',
            args: ['--disable-gpu']
        }),
                                        }
```
I include the url via "file:///", the css via string.

Results: This works perfectly fine in my scenario. 
(Besides one API change in 1.x, page.setRequestInterceptionEnabled(true) changed to await page.setRequestInterception(true)

Therefore I have created a PR (#231) for this API change, otherwise people can't use a newer version with "getBrowser"-fct.

Thanks
Simon

pocketjoso:
> I include the url via "file:///", the css via string.
> Results: This works perfectly fine in my scenario.

Well, but was the page really rendered and correct critical css returned? AFAIK the blocking issue is still there (https://github.com/GoogleChrome/puppeteer/issues/1506).

Since this seem to take ages for the Chrome team to fix (cc @addyosmani if there's anything you can do to speed this up) -what I'm considering doing is updating to the latest puppeteer in a new Penthouse version, and just clearly document that Penthouse (and critical) users must stay on the current version if they want to be able to use the file protocol with local files... it's not ideal but I too want to be able to use the latest puppeteer version.

simllll:
I have seen the open issue, but in my scenario (with latest chrome stable version) a correct critical css is returned! I have tested several situations and different pages. Generated critical css is identical to the one penthouse has generated with phantomjs. 

[...]

PR #232, opened 2018-02-21
Puppeteer 1.x support, new feature: pass html directly to puppeteer

simllll:
See #226. This would be a proposed solution.

Stability: I'm using this already in production.
Full backwards compatible to older puppeteers.

When the file:// problem gets fixed and you want to bypass the workaround, just emove line 46 to 50 from core.js.

Regards
Simon


pocketjoso:
@simllll I will close this PR now, because #226 solved the most important upgrade issue, and because this PR as it stands would need significant rebasing to move forward. Let me know if you think there are remaining changes here that carry enough value to be worked on again.

simllll:
Hi Jonas,
thanks for the update. I'm heavily using the "pass html as string" feature that is used in my patch. I can write a new patch for the current master version if you think that it gets a chance to get merged.

Basically it's a feature request that allows ,besides passing any url or file, also a string that cointains the HTML of the page in it. But there is actually a workarround available by passing something like this as url: `data:text/html;charset=UTF-8,<html>:....</html>`
need to do some testing, could be that this is sufficient for my use case and a patch is not needed.

regards

pocketjoso:
I see. Well test if your workaround is sufficient, if no I agree no need to update Penthouse (could perhaps add an example, or mention in readme and link to a gist with the pattern).

If it isn't then I would be OK with accepting a _new_ pr just for that feature. I think it would be cleaner to have a new prop though, `htmlString`, to match the `cssString` property that already exists.

PR #233, opened 2018-02-22
Move Jest to devDependencies

topaxi:
I noticed that one of my projects where I use penthouse pulled in a lot more dependencies, this was due to penthouse depending on Jest. I assume this project only depends on Jest in development/testing?

Feel free to close this PR if this is not the case.


Saturate:
LGTM!

But do note that this also bumps the Jest version.

pocketjoso:
Thanks for the PR!

The bump should be fine; the tests still succeed, and I double checked their changelogs and nothing worrying there.

PR #240, opened 2018-03-04
improve parallelization by minimizing await usage

pocketjoso:
mostly has an impact when taking before/after screenshots, and when the page load is not being the bottle neck. In my production tests the latter is most often the bottle neck however, so hence why this code is better in theory, in practice it does not make much of a difference. I think it is slightly cleaner though.


pocketjoso:
Little bit stuck unexpected errors in tests after these updates, seemingly due to a race condition bug in headless chrome. Will try updating to land latest puppeteer version next to see if that also fixes this issue.

PR #244, opened 2018-03-07
puppeteer@1.0.0

pocketjoso:
cannot go to latest version (1.1.1 at time of writing this)
because it reintroduces the `file://` protocol bug that 1.0.0 fixes:

https://github.com/GoogleChrome/puppeteer/issues/1506#issuecomment-371227988

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "css-tree": "1.0.0-alpha.27",
     "debug": "^3.1.0",
     "jsesc": "^1.0.0",
-    "puppeteer": "0.12"
+    "puppeteer": "1.0.0"
   },
   "devDependencies": {
     "babel-cli": "^6.26.0",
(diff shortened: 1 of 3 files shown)


bezoerb:
\o/ 
Just tested the current master against my bump-dependencies branch.
It seems this update fixes the ECONRESET error over at critical :)

pocketjoso:
Good! :) I'll cut a new release of Penthouse in the next few days.
Originally wanted to take the latest version of puppeteer and some related
changes, but will do an intermediate release until that's possible.

pocketjoso:
@bezoerb i have been thinking a bit mkre about how users call penthouse
with multiple urls, often in parallel, and how it can cause issues that
users don't understand.. I created a new examples folder to help educate a
little bit, but I was wondering:

If penthouse added a urls field (plural), accepting an array, and taking
care of simply putting a cap at max be of jobs being requested in parallel
(simple queue).. would that be of any help to critical package? How is it
handled in your end when users specifies multiple dimensions for example?
Executed in sequence? All at the same time? Or some kind of queue?

pocketjoso:
@bezoerb cut a new release now, 1.4.1:
https://github.com/pocketjoso/penthouse/releases/tag/v.1.4.1

PR #248, opened 2018-03-21
fix example, CSS uses semicolon instead of colon.

joejordanbrown:


pocketjoso:
Great, thanks for the PR!

PR #251, opened 2018-03-28
fixed error "Execution context was destroye

hummal:


hummal:
Iam not sure why there are all the other commits. I will try again.

PR #252, opened 2018-03-28
fixed error "Execution context was destroyed"

hummal:
This is the PR to fix the execution context was destroyed problem of https://github.com/pocketjoso/penthouse/issues/202

### DIFF
--- a/src/browser-sandbox/pruneNonCriticalSelectors.js
+++ b/src/browser-sandbox/pruneNonCriticalSelectors.js
@@ -90,12 +90,17 @@ export default function pruneNonCriticalSelectors ({
     )
   }
 
-  // give some time (renderWaitTime) for sites like facebook that build their page dynamically,
-  // otherwise we can miss some selectors (and therefor rules)
-  // --tradeoff here: if site is too slow with dynamic content,
-  // it doesn't deserve to be in critical path.
-  return sleep(renderWaitTime).then(() => {
-    console.log('debug: waited for renderWaitTime: ' + renderWaitTime)
-    return filterSelectors(selectors)
-  })
+  try {
+    // give some time (renderWaitTime) for sites like facebook that build their page dynamically,
+    // otherwise we can miss some selectors (and therefor rules)
+    // --tradeoff here: if site is too slow with dynamic content,
+    // it doesn't deserve to be in critical path.
+    return sleep(renderWaitTime).then(() => {
+      console.log('debug: waited for renderWaitTime: ' + renderWaitTime)
+      return filterSelectors(selectors)
+    })
+  } catch (err) {
+    console.log('debug: ' + err.message)
+    return []
+  }
 }
--- a/src/core.js
+++ b/src/core.js
@@ -24,46 +24,31 @@ async function loadPage (page, url, timeout, pageLoadSkipTimeout) {
   let waitingForPageLoad = true
   let pageLoadSkipTimeoutPromise = null
   const loadPagePromise = new Promise((resolve, reject) => {
-    page
-      .goto(url, {
-        timeout: timeout + 1000
-      })
-      .then(response => {
-        // Reject exits the Promise.all call immediately and ensures that the pageload is always higher prio than timout
-        return resolve('loadPageResponse')
-      })
-      .catch(err => {
-        // Reject when error because we don't want an errored page
-        console.error(err)
-        return reject(err)
-      })
+    page.goto(url, {
+      timeout: timeout + 1000
+    })
   })
 
   if (pageLoadSkipTimeout) {
     pageLoadSkipTimeoutPromise = new Promise((resolve, reject) => {
       // enables us to control requests(stop, continue, modify)
-      page.on('response', async response => {
-        if (response.url === url) {
-          debuglog('RESPONSE URL: ' + response.url) // TODO: interceptedRequest.url is a function in puppeteer > 0.12
-          if (pageLoadSkipTimeout) {
-            debuglog('pageLoadSkipTimeout injected on dom creation')
-            // instead we manually _abort_ page load after X time,
-            // in order to deal with spammy pages that keep sending non-critical requests
-            // (tracking etc), which would otherwise never load.
-            // With JS disabled it just shouldn't take that many seconds to load what's needed
-            // for critical viewport.
-            setTimeout(() => {
-              if (waitingForPageLoad) {
-                debuglog(
-                  'page load waiting ABORTED after ' +
-                    pageLoadSkipTimeout / 1000 +
-                    's. '
-                )
-                resolve()
-              }
-            }, pageLoadSkipTimeout)
+      page.on('response', response => {
+        debuglog('pageLoadSkipTimeout injected on dom creation')
+        // instead we manually _abort_ page load after X time,
+        // in order to deal with spammy pages that keep sending non-critical requests
+        // (tracking etc), which would otherwise never load.
+        // With JS disabled it just shouldn't take that many seconds to load what's needed
+        // for critical viewport.
+        setTimeout(() => {
+          if (waitingForPageLoad) {
+            debuglog(
+              'page load waiting ABORTED after ' +
+                pageLoadSkipTimeout / 1000 +
+                's. '
+            )
+            resolve()
           }
-        }
+        }, pageLoadSkipTimeout)
       })
     })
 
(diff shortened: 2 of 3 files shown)


src/core.js
```diff
@@ -23,32 +22,65 @@ async function loadPage (page, url, timeout, pageLoadSkipTimeout) {
...
+      })
+      .then(response => {
+        // Reject exits the Promise.all call immediately and ensures that the pageload is always higher prio than timout
+        return resolve('loadPageResponse')
+      })
+      .catch(err => {
+        // Reject when error because we don't want an errored page
+        console.error(err)
+        return reject(err)
+      })
```
pocketjoso: I'm not sure this `then...catch` code does anything, but tell me if you think I'm wrong! The error would have rejected the promise by itself, so catching and manually rejecting it should be redundant. Similarly, you are not making use of the explicit `loadPageResponse` value that you added, hence that `thenable` seem unnecessary.

src/core.js
```diff
@@ -23,32 +22,65 @@ async function loadPage (page, url, timeout, pageLoadSkipTimeout) {
...
-          if (waitingForPageLoad) {
-            debuglog(
-              'page load waiting ABORTED after ' +
-                pageLoadSkipTimeout / 1000 +
-                's. '
-            )
-            resolve()
+    pageLoadSkipTimeoutPromise = new Promise((resolve, reject) => {
+      // enables us to control requests(stop, continue, modify)
+      page.on('response', async response => {
```
pocketjoso: I don't understand this logic, and what it has to do with the pageLoadSkipTimeout... that timeout is meant to run in parallel with the page load, to exit if the page load takes to long. Moving it into a response handler defeats that purpose.

src/core.js
```diff
@@ -23,32 +22,65 @@ async function loadPage (page, url, timeout, pageLoadSkipTimeout) {
...
+        if (response.url === url) {
```
pocketjoso: what are you trying to do here? Why is it relevant that we got a response for the right url? As far as I can see it's not relevant for the pageLoadSkipTimeout, so what it is relevant for?

src/core.js
```diff
@@ -330,6 +367,7 @@ async function pruneNonCriticalCssLauncher ({
     // -> [BLOCK FOR] critical css selector pruning (in browser)
     let criticalSelectors
     try {
+      await page.waitFor(300) // This is needed to  fix the Target closed error of puppeteer
```
pocketjoso: I would still prefer for this to be a prop. You asked who cares, and as I've mentioned before I run tens of thousands of penthouse jobs per day, so I do care (about this, and penthouse execution timings in general). :)

src/core.js
hummal: You are right. This was a leftover of my previous version. Will change it later on

src/core.js
hummal: this is also a leftover. I tried to find out why the error is occuring. So I just want to start the timeout when the page original url was requested and the response came in. In my opinion this is the real pageload. In this case we are talking about pageload and not response time. Response time is what your code measures.

src/core.js
hummal: I cant tell you why this is needed to be an wrapped promise. But I got different errors regarding a closed target. 
You can remember my investigation ins this part of the code where the response size was different when the pageLoadSkipTimeout was close the the real page load timer. This complete section solved that problem.

src/core.js
hummal: My error case wasn't reported just by me. I read it 2 or 3 times in your issues list. So I guess you shouldn't ignore it. 
I can do it with an additional property. That is not a problem, but I would prefer a stable software which I can rely on. It doesn't matter if code runs 300ms faster per execution when there is a chance of getting errored results. 
But if you wish I will create a prop for that case.

hummal:
> Are these changes independent? Is the waitFor change enough to fix your problem?
unfortunately not. I tried to reduce to code as much as possible. But the page load promise is an important part of it. It is not exaclty the same error, but there is a chance of getting the target closed when the timeout and timer getting close before resolving and then there is a 499 status code where the client closed the connection. and the document is not completely transfered. This is absolutely not what you want. :)

> I wonder if the problem you raised in #256 messed up with what you want to test.. Probably not, but to be sure I would like to fix that issue first. As you didn't that code specifically in this PR, I might open up a separate PR just to fix that.

I followed you wished and didn't address different things in one PR. So I thought it would be best to create an issue. you check it and then we talk about fixing it.

src/browser-sandbox/pruneNonCriticalSelectors.js
```diff
@@ -90,12 +90,17 @@ export default function pruneNonCriticalSelectors ({
...
 
-  // give some time (renderWaitTime) for sites like facebook that build their page dynamically,
-  // otherwise we can miss some selectors (and therefor rules)
-  // --tradeoff here: if site is too slow with dynamic content,
-  // it doesn't deserve to be in critical path.
-  return sleep(renderWaitTime).then(() => {
-    console.log('debug: waited for renderWaitTime: ' + renderWaitTime)
-    return filterSelectors(selectors)
-  })
+  try {
```
pocketjoso: This added try/carch should not be necessary; the call of this function is already wrapped in a try/catch in core.js.

pocketjoso:
>> I wonder if the problem you raised in #256 messed [...]

> I followed you wished and didn't address different things in one PR. So I thought it would be best to create an issue. you check it and then we talk about fixing it.

Thanks for that - I merged the change into master.

pocketjoso:
As for the changes in this PR, they still don't seem like obvious good fixes to me, I'm still hesitant to merge them.. sorry. I'm looking at what we can do, so I read up further to try to understand the problem better, via this thread (as was posted earlier in the issue): 
https://www.bountysource.com/issues/51332369-when-would-an-error-cannot-find-context-with-specified-id-undefined-happen
I mostly agree with this comment (From MatthewMueller)
> The underlying problem is that there is a race condition between the event that clears execution contexts and the event that adds new execution contexts.
When you navigate to a new page, sometimes headless chrome will add the new contexts, before clearing the old ones. When it goes to clear the contexts, it blows away the recently added context, causing: Cannot find context with specified id undefined.
I think the fix should come from upstream, since I think it's going to be quite difficult to handle correctly in puppeteer.

So, if I understand it the only contribution for the original issue here would be the timeout, which is a crude and arbitrary workaround for the problem. Ideally I would like a cleaner solution.

1. did you check if you still have the issue with (`master` Penthouse and) the latest version of Puppeteer (as long as you're not running on localhost, as that does not work on that version)?

If that does not work did you try any of the other solutions mentioned in that tried, which I think are cleaner:
(if you implement either 2 or 3 - note that in the code they belong in the loadPage function, in a thenable appended to the `page.goto`. We want the page to really be loaded before this function resolves, which it currently appearently is not. We still want to keep the `pageLoadSkipTimeout` promise separate, as it serves a different purpose)

2.
```js
await page.waitForSelector('html')
```
If that works it's the simplest and cleanest workaround I saw (as it should always work, and should only wait for as long as necessary, rather than an arbitrary time).

3. A solution like `wait-for-navigation-and-context.js`
From the [same page](https://www.bountysource.com/issues/51332369-when-would-an-error-cannot-find-context-with-specified-id-undefined-happen)
This is a bit more involved, and I do wonder if it wouldn't cause the same crash.. but if it does work, I think it's a better workaround then an arbitrary timeout.

hummal:
> 2. `await page.waitForSelector('html')`

Didn't work out. Still got error. With waitFor(500) its not there. This is the reason why I chose this. And in my case the browser is crashing a lot of times and we still have unhandled promise rejections. 
Even if I try to waitFor body or a div selectors it is not working as expected. 

Generally: I understand your purpose but I can't go this way with you. This software needs this workaround and I don't have so much time to see my prs going to not be included. 
Iam going to work on my own puppeteer wrapper. Thanks for inspiration and sorry for being less patient but time is money in my case.

pocketjoso:
@hummal just use a fork of Penthouse with the added `waitFor` time, until this problem is fixed upstream? isn't that easier than creation your own wrapper from scratch?

hummal:
@pocketjoso I already do, but it feels wrong to use a fork of an active developed repo. I will write one for my business clients which rely on a successful output without errors. I can't wait for them to fix it. And maybe I will get around this error. Thanks anyway

PR #260, opened 2018-05-09
Do not recreate browser instance if getBrowser options passed

sunify:
Without this check, several browser instances are created. It leads to the process hanging.


pocketjoso:
Thanks for the fix! 👍

sunify:
Wow, so fast. Thanks!

PR #261, opened 2018-05-13
Puppeteer latest

pocketjoso:
chromium fixed bug causing file:// protocol to not be properly supported, and fix now in Puppeteer, so this updated to the latest version, `puppeteer@1.4.0` 🎉 

Also upgraded other deps.


pocketjoso:
Available as `penthouse@1.4.5`.

PR #272, opened 2018-09-02
Timeout fixes

pocketjoso:
Since Penthouse@1.8 there has been more penthouse timeouts. I traced this puppeteer never firing the load event for sites with https errors. Eventually I got a tip on a workaround for this (https://github.com/GoogleChrome/puppeteer/issues/3118#issuecomment-417754246), so should work better now. This workaround should also resolve the blocker for upgrading puppeteer (to 1.7), but I will test that later - one thing at the time.


pocketjoso:
Works well.

PR #279, opened 2018-10-28
internal job counter determining if should close browser

pocketjoso:
The previous logic could lead to race conditions; the browser instance a later penthouse call was relying on could be closed by an earlier penthouse execution. This PR fixes that. This logic has had bugs before in certain edge cases so I will leave this PR open for a while and do further tests before merging. I will publish a release candidate of it.


pocketjoso:
Now testing as `penthouse@1.10.1-rc`.

Issue #284, opened 2019-02-08
No error on a 401 response caused by failed basic http auth.


dr0bz:
Hello @pocketjoso ,

penthouse is not throwing an error if the server responds with 401. Penthouse is running against a basic http auth protected url. If the credentials in the url (https://foo:bar@www.example.com) are wrong or just missing - penthouse is not throwing an error and completes successfully.

Penthouse version: 1.10.2

Here is an example of a basic auth protected url for testing: 

http://www.httpwatch.com/httpgallery/authentication/authenticatedimage/default.aspx?0.7349707232788205

Credentials: httpwatch:whatever

Best regards

pocketjoso:
All right, thanks for the bug report. If you have time, feel free to take a look at opening up a PR with a fix with this. If so, just let me know when you start working on it too - so we can avoid double work. Otherwise I'll get around to it eventually.

dr0bz:
Hey @pocketjoso ,

fine, i'm looking into. Perhaps I can solve it.

Regards

dr0bz:
Thank you for the support!

dr0bz:
Do you plan to make a new release? I would like to use the new feature directly.

Regards,
dr0bz

pocketjoso:
Released as `penthouse@2.1.0`. Will write release notes later, but this is the only change.

dr0bz:
Thank you!

PR #285, opened 2019-02-09
#284: throw an error on server error response

dr0bz:
Penthouse isn't stopping nor is throwing an error when the server is responding with an error status code. With this change only 200-299 response status codes will let penthouse proceed.

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -9,7 +9,8 @@ import url from 'url'
 
 import normaliseCss from './util/normaliseCss'
 
-const serverPort = 8888;
+const serverPort = 8888
+const allowedResponseCodeError = /Server response status/
 
 function staticServerFileUrl (file) {
   return 'file://' + path.join(process.env.PWD, 'test', 'static-server', file)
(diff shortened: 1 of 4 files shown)


pocketjoso:
Allright, looks good overall. However I thought about this a bit more, and to avoid introducing a breaking change _and_ to support all existing use cases (imagine someone wanting to create critical css for their 404 page), I suggest we refactor this a bit to introduce this through a new option. If you're up for it I'll let you decide how to model it, my criteria are just:
- the default value should work as today; i.e. doesn't throw on _any_ status code
- easy to configure to only accept `2xx` status codes, or `2xx` + `404`

OK for you?

test/basic.test.js
```diff
@@ -4,17 +4,39 @@ import path from 'path'
...
 
 import normaliseCss from './util/normaliseCss'
 
+const serverPort = 8888;
+
 function staticServerFileUrl (file) {
   return 'file://' + path.join(process.env.PWD, 'test', 'static-server', file)
 }
 
+function responseSatusUrl (code) {
```
pocketjoso: small typo here in function name.

dr0bz: Yes, i'll fix that.

dr0bz:
Yes, I fully agree with you: it should be optional. Disallowing 3xx responses could also break things. 

> or `2xx` + `404`

Do you mean here it could be a mix of codes?

I would suggest an option name `acceptedResponseCode` with following value types `string|number`. 
- String value: `success`
  - means all `2xx` codes
- Number value: concrete http response status code for ex. `200`

What do you think?

pocketjoso:
I think a rehex would be more powerful, it would be easy to handle all
cases. Alternatively it could also be a callback function where either the
status code or the full response is provided, and the return value should
be true or false to signify whether the response is ok to proceed with or
not. Wdyt?

Le dim. 10 févr. 2019 14:00, Hermann Bier <notifications@github.com> a
écrit :

pocketjoso:
friendly ping, if you want to have this merged please check my last comment. :)

dr0bz:
Hi @pocketjoso,

sorry for the delay. Your suggestion regarding regex/function is fine. What about to allow all of the option values:

`allowedResponseCode`: `number|regex|function` ?

Regards,
dr0bz

pocketjoso:
Wonder if the number is really useful, as with a single number you would only be able to support exactly the `200` status, not any ranges or combinations of statuses. In the end I'll leave it up to you, but I think `regex|function` might be enough.

dr0bz:
I've implemented the number option as well. I think the number option is handy, if you expect a particular status code.

Best regards,
dr0bz

PR #287, opened 2019-03-29
feat: add htmlString option

simllll:
since https://github.com/GoogleChrome/puppeteer/issues/907 is closed, it is possible to use setContent in a similar way. And as it seems that my workarund with "data:text/html;charset=UTF-8,<HTML CONTENT>" as url paremter stopped working with the latest updates, I looked for a way to integrate this with less effort.

By checking the "url" paremeter if it starts with data:text/html, we can easily check if we should use "setContent" or "goto".

All the best,
Simon


pocketjoso:
Hi again, looks like we should be able to land the feature this time without extensive changes to the library :). However I have the same questions as last time, looking at the existing props:

Name | Type | Default | Description
-- | -- | -- | --
url | string |   | Accessible url. Use file:///protocol for local html files.
cssString | string |   | Original css to extract critical css from
css | string |   | Path to original css file on disk (if using instead of cssString)

1. Can you confirm that `file://` protocol doesn't work for your use case, and if so, why? I presume this is the case, but just checking as otherwise no need to make any change.
2. If needed then, can you just add a new prop called `htmlString` instead of overloading the `url`? This will match what's done for `css`.

Then should be good to merge! 👍

simllll:
Thanks for your quick feedback,
1.) Yes, we are not able to use file://. Our usecase is that we "hijack" the express render function and only get the raw html of the currently rendered page. We cannot write this content to a temporary file because it is also required that this "document" is actually able to read files from the web (e.g. access to http, and no access to file://) to get rendered correctly.

2.) just commited the requested changes! also added a test and updated readme.md. let me know :)

Regards
Simon

pocketjoso:
Looks good - I will run some tests this weekend, but should be good to merge (after).

pocketjoso:
Allright, I've done a first batch of testing. This update is currently available as:
`penthouse@2.1.0-rc`, if you want to try it as well.

I was thinking about regarding what you wrote in the readme - the fact that the `htmlString` option will not work if the html contains any relative (asset) links. I did some tests myself and this can be mitigated by adding a [base](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) element with the proper url - if it's known. I don't really want to put html parsing into penthouse, but it could be worth mentioning in the readme. It's a bit easier to add such an element than to rewrite all relative paths. And on this note I'm just curious - what about for your use case, how did you handle this? No relative asset pats?

Finally, we've discussed how assets loading is affected by using htmlString - and never actually navigating to the real URL. I do wonder if it breaks anything else - like JS execution? since we're not on the original URL origin, I can imagine certain code executing running into errors when using `htmlString`.. but perhaps OK to release as is and just document better later when we have concrete examples of this.

pocketjoso:
⚠ Seems like this change causes Penthouse to use a lot more memory (when executing a lot of penthouse calls, in parallell and over time).. I'm looking into why this could be. The only suspects in this code I can think of would be:
`page.setContent`
and the `throw` you added in the `loadPage` function. I don't really see why either of these would consume a lot more memory, but yeah.. I'll share what I can find.

For your own use case this might not matter, and you could perhaps just use `penthouse@2.1.0-rc`. I won't merge this until this problem has been resolved.

pocketjoso:
Not sure why, but seems like it's the `page.setContent` call that causes memory usage to go up, significantly. It's just a bit unexpected to me.

simllll:
Interesting, thanks for all the details and the quick rc! Maybe you discovered an issue with puppeteer, I will also look into this on my environment next week, what usage do you see? How much "bigger" is the memory usage in your case in comparison to the go-to page method?

For the relative urls, we have two cdns for all assets, one with thumbor in front for dynamic image sizes and another for all static assets including css and JavaScript. Therefore all assets have absolutel urls anyways.
For js: till today I didn't run into any issues with the missing origin, but your are right that this could be relevant for other cases!

pocketjoso:
Yeah possibly a puppeteer issue. For me with ~10 penthouse calls running concurrently using the htmlString option roughly like 100kb html each, I see an increase in memory usage of over 1GB. If you run few penthouse jobs in parallel or have lots of RAM available this probably isn't very noticable for you.

Thanks for the info regarding your use!

simllll:
Any news on this? It's not in 2.1 as far as I can tell.

simllll:
I've fixed the PR, I cannot upgrade to 2.1 right now, as I was using the 2.1-rc1 with this patch included. What are you current thoughts about this, do you have plans to merge it in a following release?

pocketjoso:
I'll prepare a new release candidate with this as soon as I have a moment, so I can monitor if it still causes increased memory usage with the latest puppeteer version.
This way you will also be able to use that release candidate in the meantime. I'm still a bit hesitant merging it into master if it doesn't perform well - let's wait for the results.

test/core.test.js
```diff
@@ -27,6 +27,26 @@ describe('penthouse core tests', () => {
       })
   })
 
+  it('should match exactly the css in the yeoman test via setContent methode', () => {
+    var yeomanFullHtmlFilePath = path.join(process.env.PWD, 'test', 'static-server', 'yeoman.html')
+    var yeomanFullCssFilePath = path.join(process.env.PWD, 'test', 'static-server', 'yeoman-full.css')
+    var yeomanExpectedCssFilePath = path.join(process.env.PWD, 'test', 'static-server', 'yeoman-medium--expected.css')
+    var yeomanCss = read(yeomanFullCssFilePath).toString()
+    var yeomanExpectedCss = read(yeomanExpectedCssFilePath).toString()
+    var yeomonHTML = read(yeomanFullHtmlFilePath).toString()
```
pocketjoso: typo (yeoman)

test/core.test.js
```diff
@@ -27,6 +27,26 @@ describe('penthouse core tests', () => {
       })
   })
 
+  it('should match exactly the css in the yeoman test via setContent methode', () => {
```
pocketjoso: typo (method)

pocketjoso:
Published another release candidate now, will run tests on my end again:
`penthouse@2.1.1-rc2`

pocketjoso:
Just tested and still the same: leads to very high Disk I/O and memory usage.

You said you tested it before. Can you share what config you used for penthouse? Maybe you let the puppeteer browser penthouse uses get killed in between each run (default settings), which would probably reset things and avoid this problem. When I'm testing it I am managing the browser instance myself, keeping it alive.

simllll:
How do you run the tests? Are you talking about the unstableKeepBrowserAlive flag?

pocketjoso:
> How do you run the tests? Are you talking about the unstableKeepBrowserAlive flag?
Yes. What I do is I let my criticalcss.com service run a portion of it's traffic with the release candidate in question (yes, using `unstableKeepBrowserAlive`) and then I monitor resource usage. When I tried it today, just as before, I immediately saw a huge increase in Disk I/O and memory usage, which returned to normal after I revert the change.

simllll:
Thanks for your quick response again, in the meantime I looked closer and also did some library updates to solve some warnings I got.
- puppeteer 1.15
- babel to babel 7
- fixed a mix up between eslint: prettier and standard can work nicely together if you use "prettier-standard".

regarding the performace issue:
I also added a 2nd perf test for the htmlString variation. I cannot find any differences in the test, except that the setContent method is actually a little bit faster than the file method (even though I had to increase the liit to 15 seconds, due to the reason that both run around 8 seconds on my device.. is my laptop too old? ;D)

check out perf.htmlstring.test.js

regards
simon

simllll:
As a side node, instead of keeping the browser alife within the node process, it could be worth looking into https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerconnectoptions and start chrome seperately as a service. 

see also this thread: https://stackoverflow.com/questions/52431775/whats-the-performance-difference-of-puppeteer-launch-versus-puppeteer-connect

pocketjoso:
Thanks for the updates. The puppeteer upgrade I already have on another branch so it will probably be merged before this PR, but the other upgrades are nice - I might cherry-pick them into the master branch.

The updated test is unfortunately not tremendously useful and this is just because as I said before the existing test is not very useful. In particular neither test will catch the problems I am observing here, as they appear over time.

> do you see the memory increase even though you are actually not using the setContent..

No, I used this branch _without_ the `htmlString` option and it works fine. It is just when using htmlString that the problem appears. It seems like the increase in I/O activity and memory comes from puppeteer, but I don't know for sure at this point, I haven't looked at it closely.

> we need to find a way to reproduce this behavior in a test to see the real cause of this I guess.

I agree. I think this will need a test using the `unstableKeepBrowserAlive` flag, running something like 100 penthouse jobs, with at least some concurrency, something like 5 penthouse jobs run in parallel all the time. This should be sufficient to reproduce the problem.

pocketjoso:
I couldn't cherry-pick your commits from this branch, but I did merge all package upgrades into master now, via https://github.com/pocketjoso/penthouse/pull/290. Thanks for leading the way.

For this PR, at some point it need to be cleaned up - but I would't worry about that until we have:
- reproduced test case for the regression I've mentioned, with continous Penthouse use via `unstableKeepBrowserAlive` option
- a fix for this regression

I'm afraid I won't have much time to help on this, but if I find time I'll update here.

----

for your own usage, is the release candidate I pushed before OK for now? or are you blocked by this not being merged?

simllll:
I can stick to penthouse 2.1.0-rc for now.

Regarding the regression, some thoughts:
Could it be due to the networking overhead? When you use the file method, no external files can be retrieved. Using the method, all absolute referenced files can and wi be loaded. This would explain the additional memory and cpu usage, but it wouldn't actually be a bug. I will think of a method to disable external resources to load (should be doable

pocketjoso:
In my comparison I never use the "file" method. The only difference in my comparison is that in one the html comes from a string, in the other it is requested over the network. The resources wanted by the html are still requested over the network in both cases. So not sure that's what this is about.

simllll:
Oh, okay sorry then I mixed up something here ;) I will keep it in mind, maybe I can come up with another theory that is stronger ;)

Can you ensure that the additional overhead is coming from puppeteer or can it also be the node ja script itself (penthouse code)?

Regards

simllll:
Any recent development to this? I just tried to update puppeteer again, no success :(

pocketjoso:
Unfortunately not, same as before. It is blocked by that performance regression that I have shared, until there is a fix for that we can't merge. And I don't see a lot of interest for this feature so unfortunately I don't have the time to investigate myself. So for now it's stale, unless you can push it forward (reproducing the problem and coming up with a fix).

j1gs4w:
Is there any progress on this? would be really useful for my use case.

pocketjoso:
No @j1gs4w. Unfortunately this branch has also gotten a bit out of date by now. If anyone wants to take this work up, please keep in mind that in order to merge this we also need to find a way to measure and fix the regression in memory and disk I/O usage - which is not trivial.

Alternatively, I suggest using `penthouse@2.1.1-rc2`, which contains the code from the branch.

pocketjoso:
And until further I will close this PR as it's stale. We can re-open if there's any development - but I don't think I will be the one moving it forward.

PR #290, opened 2019-05-11
Upgrades

pocketjoso:
Babel 7 and eslint upgrades via @simllll and https://github.com/pocketjoso/penthouse/pull/287. Want to fix linting issue on `master` so merging here. Didn't go with `prettier-standard` in the `.eslintrc` in the end though, as it for some reason mandates not having any space between function name and parens, which I do prefer.

Puppeteer upgrade to 1.15, has been well tested at this point.


simllll:
Hi :) thanks for the update, regarding the formatting rule, there is a long discussion already on prettier itself. https://github.com/prettier/prettier/issues/1139#issuecomment-361845011

The reason why I used prettier-standard was because the rules conflicted beforehand, and probably do now still. E.g. I'm using webstorm and therefore it kept reformating the files during commit to a different style, which didn't comply with prettier afterwards. I would recommend either disabled prettier, and just use standard with e.g. eslint, or changing the style to prettiers opinionated one which has no whitespace between the function and parameters. If you like to,I can add eslint instead and make a PR?

pocketjoso:
@simllll I experienced the same problem, but for me it was fixed when I added the `indent: 0` rule in the .eslintrc:
https://github.com/pocketjoso/penthouse/pull/290/commits/f9ebbadb0dc327c5a2831c06af8962cfd4b4a9d7

PR #291, opened 2019-09-25
Properly test min-max media queries for critical CSS inclusion

therealpecus:
### What does this PR do
This PR allows penthouse to correctly keep or discard media queries bracketing a viewport interval, i.e. `@media (min-width: 320px) and (max-width: 650px)`.

Previously these MQs would be included in the css output even when not matching, regardless of the setting of `keepLargerMediaQueries`. For example, running penthouse at 650px would include a media query devised for tablets `(min-width: 768px) and (max-width: 1023px)`.

### What changed
We improved the logic that tests whether to keep or drop a media query by adding an explicit test for media queries using the AND operator. We also rewrote the logic for keepLargerMediaQueries by testing both against the current viewport size and the LLVP  (Ludicrously Large ViewPort) and keeping whatever is true.

All new logic is tested in `pre-formatting.test.js` where we added additional tests to cover the extra cases.

### Reason for this PR
We discovered this when extracting the critical CSS from a page in order to pass it to the AMP version, and having to tweak every bit to stay within the 50kb limit allowed. We noticed media queries for larger viewports were included even when not matching. It is highly debatable whether one should support AMP and furthermore try to fit a full-site-design CSS into an AMP page, but this PR fixes a minor inconsistency in the critical CSS output.

### title with emoji  👏👏👏👏 🚀🚀🚀🚀
penthouse is a terrific tool. we use it through Addy Osmani's critical wrapper and rely on it to improve FCP and FMP on all websites we code and all sites and projects we consult on. Thanks for providing it to all performance enthusiasts as we are (@madebymodo).

### DIFF
--- a/src/non-matching-media-query-remover.js
+++ b/src/non-matching-media-query-remover.js
@@ -8,7 +8,7 @@ const debuglog = debug('penthouse:preformatting:nonMatchingMediaQueryRemover')
 //  - @print
 //  - min-width > width OR min-height > height
 //    (the latter only if !keepLargerMediaQueries -- which is the default)
-//  - min-width > width AND max-width < width
+//  - min-width > width AND max-width > width
 function _isMatchingMediaQuery (mediaQuery, matchConfig) {
   // TODO: use the media query parsing from css-tree instead
   let mediaAST
--- a/src/non-matching-media-query-remover.js
+++ b/src/non-matching-media-query-remover.js
@@ -9,11 +9,7 @@ const debuglog = debug('penthouse:preformatting:nonMatchingMediaQueryRemover')
 //  - min-width > width OR min-height > height
 //    (the latter only if !keepLargerMediaQueries -- which is the default)
 //  - min-width > width AND max-width < width
-function _isMatchingMediaQuery (
-  mediaQuery,
-  matchConfig,
-  keepLargerMediaQueries
-) {
+function _isMatchingMediaQuery (mediaQuery, matchConfig) {
   // TODO: use the media query parsing from css-tree instead
   let mediaAST
   try {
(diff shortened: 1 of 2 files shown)


src/non-matching-media-query-remover.js
```diff
@@ -7,7 +7,8 @@ const debuglog = debug('penthouse:preformatting:nonMatchingMediaQueryRemover')
 // only filters out:
 //  - @print
 //  - min-width > width OR min-height > height
-// and the latter only if !keepLargerMediaQueries (which is the default)
+//    (the latter only if !keepLargerMediaQueries -- which is the default)
+//  - min-width > width AND max-width < width
```
pocketjoso: should this comment read ...`AND max-width > width`?

test/pre-formatting.test.js
```diff
@@ -26,23 +26,34 @@ process.setMaxListeners(0)
...
+      // covering combined queries
+      `@media (min-width: 320px) and (max-width: 400px) {}`,
+      `@media (min-width: 150px) and (max-width: 1700px) {}`,
     ]
+    // 1300, 1600
     const mediaToRemoveAlways = [
       `@media print {}`,
       `@media not screen {}`,
-      `@media not (min-width: 1px) {}`
+      `@media not (min-width: 1px) {}`,
```
pocketjoso: minor, but to stay in-line with the code formatting in  this repo, can you remove the trailing commas, on the last array items (everywhere in this file)? 🙏

therealpecus: sure, no problem

src/non-matching-media-query-remover.js
therealpecus: yes, definitively a typo

therealpecus:
Thanks for the code-review Jonas!

pocketjoso:
This change was just released in `penthouse@2.2.2` - thanks again!

Issue #195, opened 2017-11-03
not keyword in media queries not handles correct and causing incorrect media query removal


ryuran:
With this parameter `keepLargerMediaQueries: false` on small screen this mq `@media not screen and (min-width: 48em)` is matching but properties in it are not kept by penthouse.

When I enable the property `keepLargerMediaQueries: true`, properties in it are kept in critical CSS.

Whereas It's not a « Larger Media Query ».

I think it’s because it’s negative.

pocketjoso:
The relevant code [is here](https://github.com/pocketjoso/penthouse/blob/master/src/non-matching-media-query-remover.js#L55-L59). The default settings are to only keep the `screen` value for media type (or if the media type is omitted). Indeed there's a bug with the new `keepLargerMediaQueries` option that omits this check, so that non-matching media types are still kept.

Fixing this bug would mean that your `@media not screen` would _always_ be removed. Would that be good for your use case?

(if so, feel free to open a PR for it, or I'll get to it eventually)

ryuran:
```css
@media not screen and (min-width: 48em)
```
in css it means
```
@media not (screen and (min-width: 48em))
```
Because not always refer to the entire mq. (cf: https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries#not)

So I need content of this MQ on my critical css mobile.
Currently I can’t if I keep `keepLargerMediaQueries: true`.

I will look at it.

pocketjoso:
I had a look at the [media query parsing library](https://github.com/ericf/css-mediaquery), it looks correct.

I believe the only "bug" in the code here is `keepLargerMediaQueries` takes precedent over the other checks. I will open a PR for that now.

pocketjoso:
Seeing another bug with the `not` keyword, working on a fix.

pocketjoso:
So the PR I just opened (https://github.com/pocketjoso/penthouse/pull/200) fixes the bugs I found @ryuran. However note that the mq you listed will as I said now _always_ be removed, as the  media type value is assumed to be `screen` and hence this will never match:
```css
@media not (screen and (min-width: 48em))
```

pocketjoso:
So just to clarify, unless I'm mistaken, what you said originally is _false_:
> small screen this mq @media not screen and (min-width: 48em) is matching

ryuran:
matching by browser, but not matching by penthouse

pocketjoso:
[...]

PR #300, opened 2020-04-29
Change how inversed media queries are handled

raxbg:
Resolves issues with handling inversed media queries like the one in #195

### DIFF
--- a/src/non-matching-media-query-remover.js
+++ b/src/non-matching-media-query-remover.js
@@ -24,16 +24,17 @@ function _isMatchingMediaQuery (mediaQuery, matchConfig) {
     // rather than the inverse field, but for our purposes we want to treat
     // them the same.
     const isInverse = mq.inverse || mq.type === 'not'
-    if (
-      (!isInverse && mq.type === 'print') ||
-      (isInverse && mq.type === 'screen')
-    ) {
-      return false
-    }
     // f.e. @media all {}
     // go for false positives over false negatives,
     // i.e. accept @media randomThing {}
     if (mq.expressions.length === 0) {
+      // the checks below are only valid if the media query has no other properties other than the media type
+      if (
+        (!isInverse && mq.type === 'print') ||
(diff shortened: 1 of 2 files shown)


pocketjoso:
Hi, thanks for the PR!

Can you add a test case to verify your changes?

When I tested myself the [test case that was mentioned in the previous issue](https://github.com/pocketjoso/penthouse/issues/195#issuecomment-344868518) it still fails:
```
❌ media query should have been kept: 
    @media not all and (min-width: 50em)
```

To test this I added these two media queries mentioned in the issue into the into the test suite (https://github.com/pocketjoso/penthouse/blob/master/test/pre-formatting.test.js#L36)[here]:

Once we have a test that can demonstrate that this change works I'll merge it!

PS. when working on the test suite you have to run yarn transpile before running the test (`node_modules/.bin/jest test/pre-formatting.test.js` would be the fastest here), otherwise your changes are not taken into account. DS.

raxbg:
Hi,

Thanks for the feedback and the guidance in adding the needed tests! I just want to make sure we are on the same page regarding the `em` based query.

It looks like the `mediaToAlwaysKeep` array is used in two viewport sizes - 1300px and 1600px. Converting the example to `px` it will look like this: 
`@media not all and (min-width: 800px)` because `1em` = `16px` (see [here](https://github.com/ericf/css-mediaquery/blob/34ba6343938e72b270205447b297611c40140a4f/index.js#L171)), in which case the media should in fact be dropped instead of kept because `all and (min-width: 800px)` is `true` so inversing it means it doesn't pass the tests.

In order to make the test correct for these viewports I have used `101em`. I believe this should be a more accurate test, please let me know if I am missing anything.

pocketjoso:
Thanks for the update! You're right, I had missed that the tests were checking on viewport sizes 1300px and 1600px. 

This looks good now, will merge!

pocketjoso:
Released as `penthouse@2.2.4`.

PR #302, opened 2020-06-04
Bump puppeteer

### LINKED ISSUE #301: Please bump puppeteer
We have an issue at critical caused by puppeteer < 2.1.0.
See https://github.com/addyosmani/critical/issues/432#issuecomment-637853716

The issue seems to be fixed in puppeteer@2.1.0 so it would be nice if the used version could be bumped. Thanks :)

bezoerb:
Fixes #301

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
   "license": "MIT",
   "main": "./lib/index.js",
   "engines": {
-    "node": ">=8.0"
+    "node": ">=8.16.0"
   },
   "scripts": {
     "lint": "eslint src",
@@ -40,7 +40,7 @@
     "css-tree": "1.0.0-alpha.28",
     "debug": "^4.1.1",
     "jsesc": "^2.5.2",
-    "puppeteer": "1.15.0"
+    "puppeteer": "2.1.1"
   },
   "devDependencies": {
     "@babel/cli": "^7.6.0",


pocketjoso:
Thanks for this @bezoerb!

As it is a major bump I will do some testing before I merge this in ~1 week, as long as there are no issues. I've published a release candidate, you could use it in the meantime:
`penthouse@2.3.0-rc1`

pocketjoso:
Seeing some severe regressions with puppeteer@2, with an increase of timeouts, average job execution and incorrect yielded css.. will start debugging - but my estimate now is that this will take longer than 1 week to get merged.

bezoerb:
@pocketjoso: Thanks for looking into this. There‘s puppeteer@3 available, too. I‘ve sticked with @2 to keep the node@8 compatibility. You could try puppeteer@3 if it‘s ok to drop node@8 support ;)

pocketjoso:
Yeah I've seen pupeeteer@3 too. I do think it would be OK to drop node@8 support, don't you agree (for critical)? But my intention is to get this one merged first, then I'll follow up later looking at puppeteer@3. This will also provide you with a choice of which penthouse/puppeteer version you want to use in critical, in case you want to stick with node@8 support.

pocketjoso:
After continuing to test this change more it looks like the regressions were unrelated. I will continue to monitor for differences running with this change, but I'm back at my original estimate now: I think I'll be able to merge this sometime next week.

bezoerb:
Sounds great 😊
We already dropped node@8 support in the current version of crtitical.

pocketjoso:
Let's do this!

pocketjoso:
Released as `penthouse@2.3.0`

cco3:
puppeteer@4 is now out if you want to bump again

pocketjoso:
Thanks for the notification @cco3 - will look into it going forward.

PR #303, opened 2020-07-04
fix(browser): when usingCustomGetBrowser is set, call getBrowser each time 

simllll:
if using a custom browser, getBrowser is not called when "browser" is != null. As the behaviour of the browser instance is controlled outside, this leads to situation where a browser is closed, but penthouse believes the browser is still running.


pocketjoso:
Thanks for the fix!

(The test error is unrelated, seems to fail intermittently since the last puppeteer upgrade - need to look into at some point)

PR #305, opened 2020-08-12
interceptedRequest.url should be a function call

pyronaur:
👋 Hey!

I was going over a bug today related to URL interception and noticed that there's a bug in `test(interceptedRequest.url)` because `url` actually is going to return the function. To get the URL string the function should be called.

Here's an example from puppeteer documentation:
https://github.com/puppeteer/puppeteer/blob/main/new-docs/puppeteer.page.md#example-2


pocketjoso:
Hi - thanks for the PR! You're correct, let's get this merged! :)

pocketjoso:
Released in `penthouse@2.3.1` 🎉

PR #333, opened 2021-09-23
Ignore elements in display: none parents

letynsoft:
This PR adds ability to ignore styles for elements that are contained in hidden elements.
There is a new configuration option 'tryParentsDisplayNone' (boolean, default: false - so it doesn't change the current behavior)

### DIFF
--- a/src/browser-sandbox/pruneNonCriticalSelectors.js
+++ b/src/browser-sandbox/pruneNonCriticalSelectors.js
@@ -2,7 +2,8 @@
 // no access to scrope outside of function
 export default function pruneNonCriticalSelectors ({
   selectors,
-  maxElementsToCheckPerSelector
+  maxElementsToCheckPerSelector,
+  tryParentsDisplayNone
 }) {
   console.log('debug: pruneNonCriticalSelectors init')
   var h = window.innerHeight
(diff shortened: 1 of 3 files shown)


letynsoft:
Any news here?

pocketjoso (commented):
Thanks for the pull request and the reminder - I'm sorry about the delay, this one slipped under my radar.

Overall this look good.

One small thought:
Perhaps `excludeSelectorsWithDisplayNoneParent` is a more explicit prop name (albeit very long) - I think it's a bit better aligned with the other two props. Could you update? Can you also update the README at the same time, to document your new prop?

I'm also wondering if this might slow down the execution for pages with large DOM trees. As all nodes have the same top level parents, we're going to repeat checks on a lot of top level parent nodes.. I'm considering if it would be worth caching the response for the parents to avoid this.. but it depends on how much repetition we end up finding, and also the performance of the getComputedStyle call. I've never benchmarked this function call before, but it looks like it should be fast. So in the end, I'm OK starting like this, checking how it performs, and iterating if needed.

So if you can just update the prop name and the readme, then I'll merge this! 👍

letynsoft:
Appologies, I haven't checked this PR for quite a long time now.

I have changed the option name to the suggested `excludeSelectorsWithDisplayNoneParent` and added description into the README.md file.

PR #345, opened 2022-05-10
Fix #344 adding page goto options for puppeteer

### LINKED ISSUE #344: Navigation timeout infinitely
While trying to extract critical css for few websites we are getting navigation timed out error.

Sample Site: https://www.vision-institute.com/ 
cssString: ''body { color: red }''

tried to modify the timeout time for upto 2 minutes but it still fails.

On debugging we found out that It was getting hanged on page.goto(`site_url`)

So We checked puppeteer library and there such issues already exists. 

For now it works if we add the pageGotoOptions : reference https://github.com/puppeteer/puppeteer/blob/v14.0.0/docs/api.md#pagegotourl-options

bv-sumit:


pocketjoso:
Hi, thanks for the PR. Can you update the README.md too, to mention this new option? Then we can merge this PR.

bv-sumit:
> Hi, thanks for the PR. Can you update the README.md too, to mention this new option? Then we can merge this PR.
>> I have updated the example. Also linked the url from where the users can get all available options they can send while making request to puppeteer.

pocketjoso:
Thanks!

PR #349, opened 2022-11-14
fix uncaughtException

simllll:
sometimes it happens that viewport or interception for allowJS fails, this reuslts into an uncaughtexception. this fixes it


src/core.js
```diff
@@ -301,245 +301,250 @@ async function pruneNonCriticalCssLauncher ({
...
-    let killTimeout = null
-    async function cleanupAndExit ({ error, returnValue }) {
-      if (_hasExited) {
-        return
-      }
-      debuglog('cleanupAndExit')
-      _hasExited = true
-      clearTimeout(killTimeout)
+    // ensure there is no uncaughtException
+    try {
```
pocketjoso, line 315: Can you move the this line (the opening of the try catch block) to be below the cleanupAndExit function definition, i.e just around the code that can actually through exceptions during the runtime?

pocketjoso (commented):
Asking for a small chance, should be good to approve after.

src/core.js
simllll: (y)

pocketjoso (approved):
✅ 
Let me say for the future that I already added try/catches around like half of the code inside this new try catch - the places where I saw unhandled exceptions - and then with dedicated error handling. But given what you have found it seems that some further ones are missing, so this new try catch will help. In the future, it would be even better to instead though move it to several lower level try/catch ones with dedicated error messages, to distinguish the problems - but I'm OK with skipping that as part of this PR.

Thanks for the fix!

PR #364, opened 2025-09-28
Bump Dependencies & Convert to ESM

bezoerb:
Updates all dependencies including puppeteer

Test run through but i'm still experiencing some issues when using this at critical:
https://github.com/addyosmani/critical/pull/617

As we're already just about to have node 24 LTS i would drop the transpile step and make this ESM only.
@pocketjoso: is there any possibility to get this merged some time or should i create a new package (e.g. penthouse-esm) so that you can keep your online service running without much maintenance burden :)


bezoerb:
Testing new version in https://github.com/addyosmani/critical/pull/617

bezoerb:
As there was no response on this one i released [penthouse-esm](https://www.npmjs.com/package/penthouse-esm) now. Thanks to @pocketjoso for creating this awesome library in the first place!

